Issue #16, opened 2024-01-19
[bug]: input is cutting the connection string when saved with postgres


KevenGoncalves:
When put a string for postgres like:
`postgres://root:1234@localhost:5432/db_name?sslmode=disable`
and try to save the connection it cut the string to:
`postgres://root:1234@localhost:5432/?sslmode=disable`
and doesn't connect

KevenGoncalves:
At this moment I couldn't test postgres because of this bug

KevenGoncalves:
I have fixed this bug, it was missing the dbName path, see: #17

PR #17, opened 2024-01-19
fix(postgres): fixing postgres connection

KevenGoncalves:
This fix the problem of lazysql not opening a postgres connection because the missing dbName path 
#16


KevenGoncalves:
> The main idea of this is to be able to connect to a specific database, but this PR does not handle that scenario, it only fixes the connection url being saved with the database name but when you connect to that url, the tree stills shows every database that lives in the server.
> 
> I think this only makes sense to merge if we could handle that scenario.

yes, the tree show all the databases but it doesn't allow you to see the data or select others databases, only the one is passed in the string connection

jorgerojas26:
Yeah, but i think it would be confusing to be allowed to see other databases but not be able to interact with them. If i put a database in the connection string i would expect to only connect and see that specific database.

KevenGoncalves:
> Yeah, but i think it would be confusing to be allowed to see other databases but not be able to interact with them. If i put a database in the connection string i would expect to only connect and see that specific database.

You are right, let me fix that and open the PR again

KevenGoncalves:
@jorgerojas26  Done!.

Now only show the database that is passed in the dbName
Captura de Tela 2024-01-20 às 12 24 27

jorgerojas26:
@KevenGoncalves awesome bro. I like that.

Can you please resolve the conflicts?. I will merge this.

KevenGoncalves:
Let me try to fix it

KevenGoncalves:
@jorgerojas26 Merge conflict fixed

Issue #19, opened 2024-01-20
[improve]: Improve the colors


KevenGoncalves:
The lazysql show colors different from the normal terminal, would be great if the lazysql match the same colors like the terminal, something that happen in lazy-like TUIs like lazygit and lazydocker:

What we have:
<img width="1440" alt="Captura de Tela 2024-01-20 às 12 35 00" src="https://github.com/jorgerojas26/lazysql/assets/80492657/1fc11bde-4041-43c6-894c-b48d6d502a9d">

What is expected (the image bellow is photoshopped):
<img width="1440" alt="Captura de Tela 2024-01-20 às 12 35 00 - cópia" src="https://github.com/jorgerojas26/lazysql/assets/80492657/f46c4b5f-5510-4fff-ac47-efe3fc9326c6">

I haven't figured what I can change in the code to get this

KevenGoncalves:
Work in Progress, I could manage to get the lazysql looks like lazygit
@jorgerojas26 
<img width="1440" alt="Captura de Tela 2024-01-20 às 19 36 41" src="https://github.com/jorgerojas26/lazysql/assets/80492657/71e81749-5206-43e1-87f0-1c0ae15a0dc1">
<img width="1440" alt="Captura de Tela 2024-01-20 às 19 37 02" src="https://github.com/jorgerojas26/lazysql/assets/80492657/a0b18c31-3021-4223-a4bb-f5ae648c8a6d">

jorgerojas26:
I have a good solution for this problem.

I can open a PR with the changes, but i can wait for your solution if you want to.

KevenGoncalves:
I will open a PR so you can review it

jorgerojas26:
This has been fixed in #24

PR #22, opened 2024-01-21
Fix colors problem

KevenGoncalves:


KevenGoncalves:
#19 @jorgerojas26  check this solution for the colors problem

jorgerojas26:
Yeah, i thought about solving this the same way, but then i found a better way. I'll close this PR without merging it because i think i found a much better solution for this.

PR #24, opened 2024-01-21
Fix colors

jorgerojas26:
@KevenGoncalves Please check this solution, test it on your machine and give me some feedback.


KevenGoncalves:
Yes, I tested in my machine and works great! The only thing will be remain is to match the colors scheme to look like lazygit, Great Job!

KevenGoncalves:
I think this PR can be merge already

PR #38, opened 2024-02-29
[MYSQL] Get the primary key for the table and memoize it

itsmewes:
I'm not that familiar with Go, but thought I would take a shot at this. I have only added getting the primary key for the sql queries for MySQL to see if you are happy with how I have done it before working on any of the other drivers.

### DIFF
--- a/drivers/mysql.go
+++ b/drivers/mysql.go
@@ -13,8 +13,9 @@ import (
 )
 
 type MySQL struct {
-	Connection *sql.DB
-	Provider   string
+	Connection  *sql.DB
+	Provider    string
+	PrimaryKeys map[string]string
 }
 
 func (db *MySQL) TestConnection(urlstr string) (err error) {
@@ -55,6 +56,8 @@ func (db *MySQL) GetDatabases() ([]string, error) {
 			databases = append(databases, database)
 		}
 	}
+
+	db.PrimaryKeys = make(map[string]string, len(databases))
 
 	return databases, nil
 }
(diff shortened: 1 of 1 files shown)


jorgerojas26:
Hello, thank you for the PR.

You are not taking into account the need to pass the correct value of the primary key column. Right now you are still passing the `rowId` value which is passed as a parameter to the driver functions.

itsmewes:
> Hello, thank you for the PR.
> 
> You are not taking into account the need to pass the correct value of the primary key column. Right now you are still passing the `rowId` value which is passed as a parameter to the driver functions.

Sorry I'm not sure what you mean. This PR only takes away the hard coded `id` that was being used and replaces it with the primary key column for that specific table. The `rowId` is specified already and being used and isn't needed to get the tables primary key column. Are you able to share a snippet to show me where I need to pass in the `rowId`?

jorgerojas26:
I'm sorry for not explaining myself very well.

The before this pull request, the WHERE clause was like this:

`WHERE id = value`

Now with your PR is going to be like this:

`WHERE primaryKeyColumnName = value` where `value` is passed in the `rowId` parameter to the function.

But the `rowId` value is still being taken from the table `id` column, which is incorrect because that column may not exist.

So, while your code address the primary key column problem, the value on the WHERE clause is still a problem.

To solve that problem, every `update` or `delete` function call from mysql driver should send the `primaryKeyColumnName` value.

Sorry if i'm not explaining it very well, english is not my first language :)

itsmewes:
> I'm sorry for not explaining myself very well.
> 
> Before this pull request, the WHERE clause was like this:
> 
> `WHERE id = value`
> 
> Now with your PR is going to be like this:
> 
> `WHERE primaryKeyColumnName = value` where `value` is being passed in the `rowId` parameter to the function.
> 
> But the `rowId` value is still being taken from the table `id` column, which is incorrect because that column may not exist.
> 
> So, while your code address the primary key column problem, the value on the WHERE clause is still a problem.
> 
> To solve that problem, every `update` or `delete` function call from mysql driver should send the `primaryKeyColumnName` value.
> 
> Sorry if i'm not explaining it very well, english is not my first language :)

Thank you for going more in depth with your response, really helps. I think I don't fully grasp completely how everything works so going to take some time and get my head into the code so I can update the PR 😊

Thanks again for your help pointing me in the right direction!

jorgerojas26:
I closed this in favor of #40

PR #52, opened 2024-04-21
feat: add '?' key in order to show help informations

LinPr:
Just like lazygit, when user press '?' key, it prompt a help interface which shows the keybindings,. so I add some code the impliment the similar feature in lazysql. And after testing,  it works in lazysql now .

And also, I modify the README.md to let users know this feature.

Well, I review some of the original code and rewrite some code to make it more readable, but I did't affect any of the major function of lazysql, so it is safe to merge into the main branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ Support for multiple RDBMS is a work in progress.
 
 | Key       | Action                         |
 | --------- | ------------------------------ |
+| ?         | Help                           |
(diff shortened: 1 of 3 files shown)


LinPr:
Well?  Could you please leave some advice about what can I do to improve the code quality of closed PR, and why not want to add the feature, and what new features that I can work on.

PR #55, opened 2024-04-27
fix: add keybindings info accroding to source code

LinPr:
fix:  according to the source code , when navigate the database and table tree,, there are 2  more keys avaiable， but they are not shown in the README, so I add then in README to let user know this feature

```
tree.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
		switch event.Rune() {
		case 'G':
			childrens := tree.GetRoot().GetChildren()
			lastNode := childrens[len(childrens)-1]

			if lastNode.IsExpanded() {
				childNodes := lastNode.GetChildren()
				lastChildren := childNodes[len(childNodes)-1]
				tree.SetCurrentNode(lastChildren)
			} else {
				tree.SetCurrentNode(lastNode)
			}
		case 'g':
			tree.SetCurrentNode(rootNode)
		}
		return event
	})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ Support for multiple RDBMS is a work in progress.
 | Key | Action            |
 | --- | ----------------- |
 | L   | Focus table panel |
+| G   | Focus root tree node |
+| g   | Focus last tree node |
 
 ### SQL Editor
 
--- a/README.md
+++ b/README.md
@@ -171,11 +171,11 @@ Support for multiple RDBMS is a work in progress.
 
 ### Tree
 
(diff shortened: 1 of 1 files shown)


jorgerojas26:
While it is correct that the README doesn't document this behavior, this PR doesn't cover it well.

It should be like this:

G - Focus last root node
g - Focus first root node

So, when you hit "G" you focus the last database of the list, and when you hit "g" you focus the first one.

LinPr:
ok, I see,  that's right, i will try to fix it

Issue #54, opened 2024-04-21
Bug: program would perminately blocked when operate a empty table 


LinPr:
1. what is the problem?
> Bug: program perminately blocked when operate a empty table

2. How to reproduce this problem 
> firstly: enter into a empty table 
> secondly: press up or down key will block the total program forever, and it can not receive keyboard event anymore, even  I press ctrl+c . the program still blocked in somewhere

jorgerojas26:
Fixed in #56

PR #56, opened 2024-04-28
fix: fix perminately blocked when operate a empty table. issue: #54

LinPr:
1. fix perminately blocked when operate a empty table. issue: #54
2. replaced variable `rune  := event.Rune()` with `eventKey := event.Rune()` , since rune is a build in key word of golang, it is a build in type,   so this is not a good idea to use it as a variable .


jorgerojas26:
That's amazing, thank you for realizing that problem.

PR #58, opened 2024-06-03
Keybind support.

pnx:
This PR introduces a basic keymap system. the system has the following components:

* Commands: constants that defines an action like `MoveUp`, `Quit`. Should be used in input handlers to execute something when the command is sent.
* Bind: A struct holding a `Key` and a `Command`. This struct is used as the actual mapping between a key and command.
* Keymap: A collection of bindings. also provides a function for translating a key to a command that should be used in input handlers. the collection of bindings can be changed at runtime (loaded from config or something else in the future).

After this PR. there is 2 more functions to implement: 
* load bindings from config
* have the `?` key (this should be hardcoded) display a modal with information on what keys are binded to what commands.


jorgerojas26:
That's an awesome PR, i'll check it out a bit more and come back with an approve/suggestion.

Thanks.

PR #60, opened 2024-06-13
[Feat] Added support for external editors.

veloXm:
- Added support for **external editors** while in SQL editor.
- Users can set $SQL_EDITOR for specific editor.

- Opens another terminal for external editor.
- Users can set $SQL_TERMINAL for specific editor.

- Added **keymaps for extenal editor** in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,9 +179,13 @@ Support for multiple RDBMS is a work in progress.
 
 ### SQL Editor
 
-| Key      | Action                |
-| -------- | --------------------- |
-| CTRL + R | Run the SQL statement |
+| Key          | Action                            |
+| ------------ | --------------------------------- |
+| CTRL + R     | Run the SQL statement             |
+| CTRL + Space | Open external editor(Linux only)  |
+
+Specific editor for lazysql can be set by `$SQL_EDITOR`.
+Specific terminal for opening editor can be set by `$SQL_TERMINAL`
 
 ## Example connection URLs
 
--- a/components/SQLEditor.go
+++ b/components/SQLEditor.go
@@ -1,9 +1,12 @@
 package components
 
 import (
-	"github.com/jorgerojas26/lazysql/models"
+	"os"
+	"os/exec"
+	"runtime"
 
 	"github.com/gdamore/tcell/v2"
+	"github.com/jorgerojas26/lazysql/models"
 	"github.com/rivo/tview"
 )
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -179,9 +179,13 @@ Support for multiple RDBMS is a work in progress.
...
-| Key      | Action                |
-| -------- | --------------------- |
-| CTRL + R | Run the SQL statement |
+| Key          | Action                            |
+| ------------ | --------------------------------- |
+| CTRL + R     | Run the SQL statement             |
+| CTRL + Space | Open external editor(Linux only)  |
+
+Specific editor for lazysql can be set by `$SQL_EDITOR`.
+Specific terminal for opening editor can be set by `$SQL_TERMINAL`
```
ccoVeille, line 188: For both, I would have said

```markdown
 `SQL_WHATEVER` environment variable
```

components/SQLEditor.go
```diff
@@ -77,3 +82,95 @@ func (s *SQLEditor) SetBlur() {
...
+/*
+	THIS FUNCTION OPENS EXTERNAL EDITOR.
+
+	Q: WHY OPEN ANOTHER TERMINAL?
+	A: OPENING EDITORS LIKE VIM/NEOVIM REALLY MESSED UP INITIAL TERMINAL'S OUTPUT.
+*/
+
+func openExternalEditor(s *SQLEditor) string {
+	// Path of temporary file
+	path := "/tmp/lazysql.sql"
```
ccoVeille, line 96: Why not using os.TempFile ?

I mean it won't work when /tmp is not available, and it would always use the same file

components/SQLEditor.go
```diff
@@ -77,3 +82,95 @@ func (s *SQLEditor) SetBlur() {
...
+	}
+
+	// Read the updated content from the temporary file
+	updatedContent, err := os.ReadFile(path)
+	if err != nil {
+		return s.TextArea.GetText()
+	}
+
+	// Remove the temporary file
+	err = os.Remove(path)
```
ccoVeille, line 135: This should be moved in a defer called just after the file creation

components/SQLEditor.go
```diff
@@ -77,3 +82,95 @@ func (s *SQLEditor) SetBlur() {
...
+
+	return editor
+}
+
+// Function to select terminal
+func getTerminal() string {
+	var terminal string = os.Getenv("SQL_TERMINAL")
+
+	if terminal == "" {
+		terminal = os.Getenv("TERMINAL")
```
ccoVeille, line 168: I have never seen this variable I think

components/SQLEditor.go
```diff
@@ -77,3 +82,95 @@ func (s *SQLEditor) SetBlur() {
...
+	}
+
+	if terminal == "" {
+		terminal = "xterm"		// use "xterm" if $TERMINAL not set
```
ccoVeille, line 172: x-terninal-emulator is an alternative on Debian based system

components/SQLEditor.go
```diff
@@ -77,3 +82,95 @@ func (s *SQLEditor) SetBlur() {
...
+	content := []byte(s.TextArea.GetText())
+
+	/*
+	0644 Permission
+	* User: read & write
+	* Group: read
+	* Other: read
+	*/
+
+	err := os.WriteFile(path, content, 0644)
```
ccoVeille, line 111: The syntax is octal, it should be 0o644

README.md
veloXm: I think explicit would be better.

components/SQLEditor.go
veloXm: I have moved it to defer and will soon send a PR.

components/SQLEditor.go
veloXm: Well I've tested it on my machine and it does exist.

components/SQLEditor.go
veloXm: I have seen it earlier but decided not to include it cus (idk).
I've added it in new commit though.

components/SQLEditor.go
veloXm: I have added prefix 0, and I think that makes it a octal value.

veloXm:
I'll send a new PR soon with changes said above.

components/SQLEditor.go
ccoVeille: It does. It's valid.

The one I suggested was introduced in Go 1.13, and should be preferred if we consider some linters.

https://github.com/mvdan/gofumpt/issues/51

ccoVeille:
Thanks for reading my review. Feel free to ping me on your next PR

ccoVeille:
Found it #61

PR #67, opened 2024-07-10
fix: format table name to include backticks

### LINKED ISSUE #47: Cannot open databases with all-numeric names
Using build 0.1.8 in Debian Bookworm under WLS2

I noticed this connected to two similarly configured MySQL servers.
I do not know if it is a MySQL specific problem or not.

I can see all the databases in the table panel.
I can open ones with letters in their names, even if they start with a number.
All-numeric databases do not open. I can select them but nothing happens when I press enter.

I hope this is helpful, thanks!

jorgerojas26:
Fixes #47

### DIFF
--- a/drivers/mysql.go
+++ b/drivers/mysql.go
@@ -60,7 +60,7 @@ func (db *MySQL) GetDatabases() ([]string, error) {
 }
 
 func (db *MySQL) GetTables(database string) (map[string][]string, error) {
-	rows, err := db.Connection.Query("SHOW TABLES FROM " + database)
+	rows, err := db.Connection.Query(fmt.Sprintf("SHOW TABLES FROM `%s`", database))
 
 	tables := make(map[string][]string)
 
@@ -79,6 +79,8 @@ func (db *MySQL) GetTables(database string) (map[string][]string, error) {
 }
 
 func (db *MySQL) GetTableColumns(database, table string) (results [][]string, err error) {
+	table = db.formatTableName(table)
+
 	rows, err := db.Connection.Query("DESCRIBE " + table)
 	if err != nil {
 		return results, err
@@ -112,6 +114,8 @@ func (db *MySQL) GetTableColumns(database, table string) (results [][]string, er
 }
 
 func (db *MySQL) GetConstraints(table string) (results [][]string, err error) {
+	table = db.formatTableName(table)
+
 	splitTableString := strings.Split(table, ".")
 	database := splitTableString[0]
 	tableName := splitTableString[1]
@@ -150,6 +154,7 @@ func (db *MySQL) GetConstraints(table string) (results [][]string, err error) {
 }
 
 func (db *MySQL) GetForeignKeys(table string) (results [][]string, err error) {
+	table = db.formatTableName(table)
 	splitTableString := strings.Split(table, ".")
 	database := splitTableString[0]
 	tableName := splitTableString[1]
@@ -188,6 +193,7 @@ func (db *MySQL) GetForeignKeys(table string) (results [][]string, err error) {
 }
 
 func (db *MySQL) GetIndexes(table string) (results [][]string, err error) {
+	table = db.formatTableName(table)
 	rows, err := db.Connection.Query("SHOW INDEX FROM " + table)
 	if err != nil {
 		return results, err
@@ -218,6 +224,7 @@ func (db *MySQL) GetIndexes(table string) (results [][]string, err error) {
 }
 
 func (db *MySQL) GetRecords(table, where, sort string, offset, limit int) (paginatedResults [][]string, totalRecords int, err error) {
+	table = db.formatTableName(table)
 	defaultLimit := 300
 
 	isPaginationEnabled := offset >= 0 && limit >= 0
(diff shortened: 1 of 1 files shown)


drivers/mysql.go
```diff
@@ -313,6 +320,7 @@ func (db *MySQL) ExecuteQuery(query string) (results [][]string, err error) {
 
 // TODO: Rewrites this logic to use the primary key instead of the id
 func (db *MySQL) UpdateRecord(table, column, value, primaryKeyColumnName, primaryKeyValue string) error {
+	table = db.formatTableName(table)
 	query := fmt.Sprintf("UPDATE %s SET %s = \"%s\" WHERE %s = \"%s\"", table, column, value, primaryKeyColumnName, primaryKeyValue)
```
ccoVeille, line 324:
```suggestion
 	query := fmt.Sprintf(`UPDATE %s SET %s = "%s" WHERE %s = "%s"`, table, column, value, primaryKeyColumnName, primaryKeyValue)
```

ccoVeille, line 324: Also, I'm very worried about the  possible SQL injection

You should consider using `?` for value, then pass `value ` to Exec as an arg

drivers/mysql.go
```diff
@@ -321,6 +329,7 @@ func (db *MySQL) UpdateRecord(table, column, value, primaryKeyColumnName, primar
 
 // TODO: Rewrites this logic to use the primary key instead of the id
 func (db *MySQL) DeleteRecord(table, primaryKeyColumnName, primaryKeyValue string) error {
+	table = db.formatTableName(table)
 	query := fmt.Sprintf("DELETE FROM %s WHERE %s = \"%s\"", table, primaryKeyColumnName, primaryKeyValue)
```
ccoVeille, line 333: Same concerns here

drivers/mysql.go
```diff
@@ -385,7 +394,7 @@ func (db *MySQL) ExecutePendingChanges(changes []models.DbDmlChange, inserts []m
 
 		statementType = "DELETE FROM"
 
-		query = fmt.Sprintf("%s %s WHERE %s = \"%s\"", statementType, delete.Table, delete.PrimaryKeyColumnName, delete.PrimaryKeyValue)
+		query = fmt.Sprintf("%s %s WHERE %s = \"%s\"", statementType, db.formatTableName(delete.Table), delete.PrimaryKeyColumnName, delete.PrimaryKeyValue)
```
ccoVeille, line 397: Same

drivers/mysql.go
```diff
@@ -405,7 +414,7 @@ func (db *MySQL) ExecutePendingChanges(changes []models.DbDmlChange, inserts []m
 			}
 		}
 
-		query := fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)", insert.Table, strings.Join(insert.Columns, ", "), strings.Join(values, ", "))
+		query := fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)", db.formatTableName(insert.Table), strings.Join(insert.Columns, ", "), strings.Join(values, ", "))
```
ccoVeille, line 417: Here too the values should be passed as ?,?,?...

strings.Repeat could help

And `values` should be passed to Exec

drivers/mysql.go
```diff
@@ -442,3 +449,18 @@ func (db *MySQL) SetProvider(provider string) {
 func (db *MySQL) GetProvider() string {
 	return db.Provider
 }
+
+func (db *MySQL) formatTableName(tableName string) string {
+	splittedTableName := strings.Split(tableName, ".")
+
+	if len(splittedTableName) == 1 {
+		return tableName
```
ccoVeille, line 457: You could return this


```suggestion
		return fmt.Sprintf("`%s`", table)
```

ccoVeille (commented):
I'm sorry because most of my feedbacks are  not related to the current PR.

But I think I had to report them

jorgerojas26:
Thanks for reviewing the PR. May i ask why SQL Injection is a concern to you in this app when it is just a SQL client?.

I totally understand that i should use prepared statements, i plan to make a huge refactor on all of the queries, but i want to solve the problem with a quick solution first. Does it makes sense to you?

jorgerojas26:
The table name formatting is really dirty, but it is a quick solution. That should be refactored too.

ccoVeille:
> Thanks for reviewing the PR. May i ask why SQL Injection is a concern to you in this app when it is just a SQL client?.

I tend to always plan the worst. Often, I'm not disappointed and something close happens 😅

> I totally understand that i should use prepared statements, i plan to make a huge refactor on all of the queries, but i want to solve the problem with a quick solution first. Does it makes sense to you?

Not necessarily prepared statement, but using Exec the right way should be always preferred.

ccoVeille:
Also please consider a table could have a varchar primary key, and that this field may contain a quote

Something like `john "babayaga" wick`

It would be a bad design to rely on a varchar PK for this, but your app will be unable to work with this row

ccoVeille:
> May i ask why SQL Injection is a concern to you in this app when it is just a SQL client?

My SQL injection I was more thinking about a source of bugs, than a attack.

Maybe it's what you were asking for

jorgerojas26:
> > May i ask why SQL Injection is a concern to you in this app when it is just a SQL client?
> 
> My SQL injection I was more thinking about a source of bugs, than a attack.
> 
> Maybe it's what you were asking for

I understand. I appreciate the feedback. For now, i'll merge this PR just to resolve the connection problem and open another PR addressing your feedback.

Thank you very much.

ccoVeille:
Sure

PR #69, opened 2024-07-10
Add: help menu using '?' key

LostbBlizzard:
Capture3

Sorry it took so long I forgot. the release reminded me.

I tried making the UI similar to lazygit With the green keybind menu and blue keybinds text on the bottom of the screen and i updated the readme.

This pull request isn't really meant to be accepted right now. just checking if there's any problems with the code or UI and also I need to blue keybind text to update based on context and you can't move in the help menu using the 'j' and 'k' keys.


app/Keymap.go
```diff
@@ -38,47 +38,48 @@ func (c KeymapSystem) Resolve(event *tcell.EventKey) Command {
...
+			Bind{Key: Key{Char: '>'}, Cmd: PageNext, Description: "Next Page"},
+			Bind{Key: Key{Char: '<'}, Cmd: PagePrev, Description: "Previous page"},
 		},
 		"editor": {
-			Bind{Key: Key{Code: tcell.KeyCtrlR}, Cmd: Execute},
-			Bind{Key: Key{Code: tcell.KeyEscape}, Cmd: Quit},
-			Bind{Key: Key{Code: tcell.KeyCtrlSpace}, Cmd: OpenInExternalEditor},
+			Bind{Key: Key{Code: tcell.KeyCtrlR}, Cmd: Execute, Description: "Run Sql Statment"},
+			Bind{Key: Key{Code: tcell.KeyEscape}, Cmd: Quit, Description: "Quit"},
+			Bind{Key: Key{Code: tcell.KeyCtrlSpace}, Cmd: OpenInExternalEditor, Description: "Open extranl editer(Linux Only)"},
```
ccoVeille, line 82:
```suggestion
			Bind{Key: Key{Code: tcell.KeyCtrlSpace}, Cmd: OpenInExternalEditor, Description: "Open external editor (Linux Only)"},
```

ccoVeille:
You shouldn't add the binary, BTW

ccoVeille:
Thanks for removing the binary. Maybe it should be added to `.gitignore` file

LostbBlizzard:
> Thanks for removing the binary. Maybe it should be added to `.gitignore` file

Am not really for or against the idea. Though it was a mistake on my part

LostbBlizzard:
Just got to clean up the help status string since it a bit unreadable in the table view

jorgerojas26:
Is it possible that the modal shows up a tview.Table with borders showing the keymaps? it will make it navigable by default and i think it is prettier, but that's personal.

LostbBlizzard:
I'll try doing it

components/HelpModal.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/rivo/tview"
+)
+
+type HelpModal struct {
+	*tview.Flex
+}
+
+func NewHelpModal() *HelpModal {
+
+	//Colors
```
ccoVeille, line 17: Go comments are starting by a space.
Otherwise, they are supposed to be compiler or language instructions


```suggestion
	// Colors
```

revive can help in reporting such issues

components/Home.go
```diff
@@ -301,6 +315,30 @@ func (home *Home) homeInputCapture(event *tcell.EventKey) *tcell.EventKey {
...
+	} else if command == commands.OpenKeymapMenu {
+
+		ok := true
+		if table != nil {
+			if table.GetIsEditing() {
+				ok = false
+			}
+		}
+
+		if ok {
```
ccoVeille, line 327:
```suggestion
		if table == nil || !table.GetIsEditing() {
```

LostbBlizzard:
Just Looked over it and since I don't see any problems with it. I'll just finish feature here.

LostbBlizzard:
The last [force-pushed](https://github.com/jorgerojas26/lazysql/compare/56038bdf6cdc22e1b36639c3cf86ae0481d78317..ada8b36edd7c4dea723fb5d33e558fa321574e84) was for fixing the wrong email address on the commits.

LostbBlizzard:
I may close this pull request and make new one based on current main and readd my changes from there since. It seems easier and less likely for me to make mistake than fixing conflicts using the GitHub website UI.

PR #70, opened 2024-07-13
fixes theme colors

### LINKED ISSUE #1: Issues with Colorusage and Readability
So, i'm not sure on which end this issue is, but i'm not getting the same color output as your screenshots do
grafik

Especially the unfocused treeview is using a Blue no matter which colorscheme i use. This is in Windows Terminal, tested with 3 colorschemes.

Additionally i feel like the contrast on the selected cell is pretty low. A darker foreground color for that would help readability.

jorgerojas26:
Fixes #1

### DIFF
--- a/app/App.go
+++ b/app/App.go
@@ -8,19 +8,18 @@ import (
 var App = tview.NewApplication()
 
 func init() {
-	Styles := tview.Theme{
+	theme := tview.Theme{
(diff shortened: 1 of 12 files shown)


jorgerojas26:
The next step of this will be to allow theme customization. But for now i want to fix the colors issue which makes lazysql almost unusable on some systems.

PR #73, opened 2024-07-16
Refactor DML actions handling

jorgerojas26:
The main goal of this refactor is to improve code quality (a bit, i'm not super fluent in go) and allow for EMPTY, NULL and DEFAULT values to be used in the DML statements.

Please feel free to make suggestions.


drivers/mysql.go
```diff
@@ -352,20 +437,21 @@ func (db *MySQL) ExecuteQuery(query string) (results [][]string, err error) {
...
-// TODO: Rewrites this logic to use the primary key instead of the id
-func (db *MySQL) DeleteRecord(table, primaryKeyColumnName, primaryKeyValue string) error {
-	table = db.formatTableName(table)
-	query := fmt.Sprintf("DELETE FROM %s WHERE %s = \"%s\"", table, primaryKeyColumnName, primaryKeyValue)
-	_, err := db.Connection.Exec(query)
+func (db *MySQL) DeleteRecord(database, table, primaryKeyColumnName, primaryKeyValue string) error {
+	query := "DELETE FROM "
+	query += db.formatTableName(database, table)
+	query += fmt.Sprintf(" WHERE %s = ?", primaryKeyColumnName)
+	_, err := db.Connection.Exec(query, primaryKeyValue)
```
ccoVeille, line 454:
```suggestion
	query := fmt.Sprintf("DELETE FROM %s WHERE %s = ?", db.formatTableName(database, table), primaryKeyColumnName)
	_, err := db.Connection.Exec(query, primaryKeyValue)
```

drivers/mysql.go
```diff
@@ -384,94 +470,104 @@ func (db *MySQL) ExecuteDMLStatement(query string) (result string, err error) {
...
-
-		query := fmt.Sprintf("UPDATE %s SET %s WHERE %s = '%s';", table, updateClause, primaryKeyColumnName, primaryKeyValue)
-
-		queries = append(queries, query)
-	}
+		switch change.Type {
+		case models.DmlInsertType:
+			queryStr := "INSERT INTO "
+			queryStr += db.formatTableName(change.Database, change.Table)
+			queryStr += fmt.Sprintf(" (%s) VALUES (%s)", strings.Join(columns, ", "), strings.Join(valuesPlaceholder, ", "))
```
ccoVeille, line 507: You could/should a sprintf here to make things easier to read

See https://github.com/jorgerojas26/lazysql/pull/73#discussion_r1685676775

drivers/mysql.go
jorgerojas26: The thing is gosec G201 rule, you think it should be disabled?

ccoVeille: i think you fixed it the right way.

https://securego.io/docs/rules/g201-g202

I'm unsure if there is "clean way" when you are sprintf to set the table name

You may look at ORM how they do.

But now you understood how to write clean SQL queries with ? and arguments, maybe you could add a //nolint:gosec // disable it because of usage of table

jorgerojas26: Thanks for suggestion. I did look at how query builders like [squirrel](https://github.com/Masterminds/squirrel) do it, and definitely they are not using Sprintf, they build the query writing to a "&bytes.Buffer{}" ([see](https://github.com/Masterminds/squirrel/blob/1ded5784535dcffa4e175d4efbd1ca2706927758/insert.go#L53-L115)). 

That's a good way, maybe i'll do it.

ccoVeille: Using a buffer is another way to contact string, it doesn't help at all with SQL injection.

You can use the buffer of course, as it's fast.

I think that's why they are using it

jorgerojas26: By the time i'm more concerned about readability than SQL Injection, i think it does not makes sense (unless i'm missing something, if i do, please advice) to think that's a problem on a sql client, you will not self inject sql.

jorgerojas26: You are correct, using parameters we avoid those things.

Thank you, sir. :)

ccoVeille:
any progress? do you need help?

jorgerojas26:
I just lost my main job, i'm busy right now preparing for interviews and those things. When I have more stability I will come back and get this done. Thanks for asking.

ccoVeille:
Wow, take care of you

PR #76, opened 2024-07-17
golangci lint

### LINKED ISSUE #71: Improve code quality by using golangci-lint
We could set up golangci-lint rules in the repository that could be checked with a GitHub action.

It would help to keep code quality

ccoVeille:
- **chore: use Go canonical way to return errors**
- **chore: use the canonical name for error variable**
- **ci: add golangci-lint configuration file**
- **fix: catch errors reported by errcheck linter**
- **fix: make sure to catch all errors in drivers**
- **ci: enable more linters**
- **chore: fix import orders**
- **chore: remove dot import and use type alias**
- **chore: remove else when previous if ends with a return**
- **chore: remove type or default value when obvious**
- **chore: use go initialism for variable name**
- **chore: no uppercase constant in Go**
- **chore: remove suspicious blank imports**
- **fix: make sure to use the database name when getting table columns**
- **ci: enable golang-lint in GitHub actions**

Fixes #71 

Related to #14

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,11 @@ jobs:
       -
         name: Set up Go
         uses: actions/setup-go@v5
+
+      -
+        name: Golangci-lint
+        uses: golangci/golangci-lint-action@v6.0.1
+
       -
         name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v6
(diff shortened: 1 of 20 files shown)


.github/workflows/release.yml
```diff
@@ -20,6 +20,11 @@ jobs:
       -
         name: Set up Go
         uses: actions/setup-go@v5
+
+      -
+        name: Golangci-lint
+        uses: golangci/golangci-lint-action@v6.0.1
```
ccoVeille, line 26: I'm noticing I added this one on the release process, so only when something is tagged

PR #103, opened 2024-09-07
Add a toggleable sidebar to view records

jorgerojas26:
image

The main idea here is to have that toggleable sidebar that shows the table results in a vertical manner, so it is easy to see all the columns.

### DIFF
--- a/components/ResultsTable.go
+++ b/components/ResultsTable.go
@@ -89,9 +89,9 @@ func NewResultsTable(listOfDbChanges *[]models.DbDmlChange, tree *Tree, dbdriver
 	loadingModal.SetTextColor(tview.Styles.SecondaryTextColor)
 
 	pages := tview.NewPages()
-	pages.AddPage(tablePageName, wrapper, true, true)
-	pages.AddPage(tableErrorPageName, errorModal, true, false)
-	pages.AddPage(tableLoadingPageName, loadingModal, false, false)
+	pages.AddPage(pageNameTable, wrapper, true, true)
+	pages.AddPage(pageNameTableError, errorModal, true, false)
+	pages.AddPage(pageNameTableLoading, loadingModal, false, false)
 
 	pagination := NewPagination()
 
@@ -116,7 +116,7 @@ func NewResultsTable(listOfDbChanges *[]models.DbDmlChange, tree *Tree, dbdriver
 	table.SetFixed(1, 0)
 	table.SetInputCapture(table.tableInputCapture)
 	table.SetSelectedStyle(tcell.StyleDefault.Background(tview.Styles.SecondaryTextColor).Foreground(tview.Styles.ContrastSecondaryTextColor))
-	table.Page.AddPage(sidebarPageName, table.Sidebar, false, false)
+	table.Page.AddPage(pageNameSidebar, table.Sidebar, false, false)
 
 	table.SetSelectionChangedFunc(func(_, _ int) {
 		if table.GetShowSidebar() {
@@ -176,8 +176,8 @@ func (table *ResultsTable) WithEditor() *ResultsTable {
 	resultsInfoText.SetTextColor(tview.Styles.PrimaryTextColor)
 	resultsInfoWrapper.AddItem(resultsInfoText, 3, 0, false)
 
-	editorPages.AddPage(tableEditorTablePageName, tableWrapper, true, false)
-	editorPages.AddPage(tableEditorResultsInfoPageName, resultsInfoWrapper, true, true)
+	editorPages.AddPage(pageNameTableEditorTable, tableWrapper, true, false)
+	editorPages.AddPage(pageNameTableEditorResultsInfo, resultsInfoWrapper, true, true)
 
 	table.EditorPages = editorPages
 	table.ResultsInfo = resultsInfoText
@@ -193,7 +193,7 @@ func (table *ResultsTable) subscribeToTreeChanges() {
 	ch := table.Tree.Subscribe()
 
 	for stateChange := range ch {
-		if stateChange.Key == selectedDatabaseTree {
+		if stateChange.Key == eventTreeSelectedDatabase {
 			table.SetDatabaseName(stateChange.Value.(string))
 		}
 	}
@@ -204,13 +204,13 @@ func (table *ResultsTable) subscribeToSidebarChanges() {
 
 	for stateChange := range ch {
 		switch stateChange.Key {
-		case editingSidebar:
+		case eventSidebarEditing:
 			editing := stateChange.Value.(bool)
 			table.SetIsEditing(editing)
-		case unfocusingSidebar:
+		case eventSidebarUnfocusing:
 			App.SetFocus(table)
 			App.ForceDraw()
-		case togglingSidebar:
+		case eventSidebarToggling:
 			table.ShowSidebar(false)
 			App.ForceDraw()
 		}
@@ -493,7 +493,7 @@ func (table *ResultsTable) subscribeToFilterChanges() {
 
 	for stateChange := range ch {
 		switch stateChange.Key {
-		case filteringResultsTable:
+		case eventResultsTableFiltering:
 			if stateChange.Value != "" {
 				rows := table.FetchRecords(nil)
 
@@ -532,7 +532,7 @@ func (table *ResultsTable) subscribeToEditorChanges() {
 
 	for stateChange := range ch {
 		switch stateChange.Key {
-		case querySQLEditor:
+		case eventSqlEditorQuery:
 			query := stateChange.Value.(string)
 			if query != "" {
 				queryLower := strings.ToLower(query)
@@ -568,7 +568,7 @@ func (table *ResultsTable) subscribeToEditorChanges() {
 						}
 						table.SetLoading(false)
 					}
-					table.EditorPages.SwitchToPage(tablePageName)
+					table.EditorPages.SwitchToPage(pageNameTable)
 					App.Draw()
 				} else {
 					table.SetRecords([][]string{})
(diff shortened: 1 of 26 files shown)


components/ResultsTable.go
```diff
@@ -186,6 +199,24 @@ func (table *ResultsTable) subscribeToTreeChanges() {
...
+		case "Editing":
+			editing := stateChange.Value.(bool)
+			table.SetIsEditing(editing)
+		case "Unfocusing":
+			App.SetFocus(table)
+			App.ForceDraw()
+		case "Toggling":
+			table.ShowSidebar(false)
+			App.ForceDraw()
+		}
```
ccoVeille, line 216: Editing, Unfocusing, and Toggling should be constants

components/ResultsTable.go
```diff
@@ -1224,3 +1265,52 @@ func (table *ResultsTable) search() {
...
+			text := table.GetCell(selectedRow, i-1).Text
+			title := name
+
+			logger.Info("string repeat", map[string]any{"sidebarWidth": sidebarWidth, "name": len(name), "colType": len(colType)})
+			repeatCount := sidebarWidth - len(name) - len(colType) - 6 // 2 for spaces added below and idk why 6 is needed, but it works.
+
+			if repeatCount <= 0 {
+				repeatCount = 1
+			}
+			title += "[gray]" + strings.Repeat("-", repeatCount)
```
ccoVeille, line 1310: Shouldn't gray be a constant you define somewhere to provide a consistent theming

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	sidebar.Fields = append(sidebar.Fields, field)
+	sidebar.Flex.AddItem(field, itemFixedSize, 0, true)
+}
+
+func (sidebar *Sidebar) FocusNextField() {
+	newIndex := sidebar.GetCurrentFieldIndex() + 1
+
+	if newIndex < sidebar.Flex.GetItemCount() {
```
ccoVeille, line 82: Invert the test to simplify


```suggestion
	if newIndex >= sidebar.Flex.GetItemCount() {
		return
}
```

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+			return
+		}
+
+	}
+}
+
+func (sidebar *Sidebar) FocusPreviousField() {
+	newIndex := sidebar.GetCurrentFieldIndex() - 1
+
+	if newIndex >= 0 {
```
ccoVeille, line 98: Same check for < 0 and return

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+		item := sidebar.Fields[newIndex]
+
+		if item != nil {
```
ccoVeille, line 101: If item == nil, then return

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if newIndex < sidebar.Flex.GetItemCount() {
+		item := sidebar.Fields[newIndex]
+
+		if item != nil {
+			sidebar.SetCurrentFieldIndex(newIndex)
+			App.SetFocus(item)
+			App.ForceDraw()
+			return
+		}
```
ccoVeille, line 90: You can also simplify here

```suggestion
		if item != nil {
			return
		}
		sidebar.SetCurrentFieldIndex(newIndex)
		App.SetFocus(item)
		App.ForceDraw()
```

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+
+	sidebar.SetEditingStyles(item)
+}
+
+func (sidebar *Sidebar) inputCapture(event *tcell.EventKey) *tcell.EventKey {
+	command := app.Keymaps.Group(app.SidebarGroup).Resolve(event)
+
+	switch command {
+	case commands.UnfocusSidebar:
+		sidebar.Publish(models.StateChange{Key: "Unfocusing", Value: nil})
```
ccoVeille, line 143: Here is an example of where you should reuse the constants I was talking about

https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749229328

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+	case commands.ToggleSidebar:
+		sidebar.Publish(models.StateChange{Key: "Toggling", Value: nil})
```
ccoVeille, line 145: See https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749231396

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+	case commands.MoveDown:
+		sidebar.FocusNextField()
+	case commands.MoveUp:
+		sidebar.FocusPreviousField()
+	case commands.GotoStart:
+		sidebar.FocusFirstField()
+	case commands.GotoEnd:
+		sidebar.FocusLastField()
+	case commands.Edit:
+		sidebar.Publish(models.StateChange{Key: "Editing", Value: true})
```
ccoVeille, line 155: See https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749231396

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+		text := item.GetText()
+
+		sidebar.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
+			command := app.Keymaps.Group(app.SidebarGroup).Resolve(event)
+
+			switch command {
+			case commands.CommitEdit:
+				sidebar.SetInputCapture(sidebar.inputCapture)
+				sidebar.SetDisabledStyles(item)
+				sidebar.Publish(models.StateChange{Key: "Editing", Value: false})
```
ccoVeille, line 168: See https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749231396

components/Sidebar.go
```diff
@@ -0,0 +1,230 @@
...
+				return nil
+			case commands.DiscardEdit:
+				sidebar.SetInputCapture(sidebar.inputCapture)
+				sidebar.SetDisabledStyles(item)
+				item.SetText(text, true)
+				sidebar.Publish(models.StateChange{Key: "Editing", Value: false})
```
ccoVeille, line 174: See https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749231396

components/constants.go
```diff
@@ -5,6 +5,7 @@ import "github.com/jorgerojas26/lazysql/app"
 var App = app.App
 
 const (
-	EditorTabName string = "Editor"
-	HelpPageName  string = "Help"
+	EditorTabName   string = "Editor"
+	HelpPageName    string = "Help"
+	SidebarPageName string = "Sidebar"
```
ccoVeille, line 10: Unless you think your constant are reused externally, they should be unexported.

So this would be better

```suggestion
	editorTabName   string = "Editor"
	helpPageName    string = "Help"
	sidebarPageName string = "Sidebar"
```

also usually constant of the same kind share a common prefix, not suffix. this way you can use the first letter to find them all in your IDE

So this would be what I would use


```suggestion
	tabNameEditor   string = "Editor"
	pageNameHelp    string = "Help"
	pageNameSidebar string = "Sidebar"
```

ccoVeille:
I'm glad you accepted my suggestion 👍

jorgerojas26:
> I'm glad you accepted my suggestion 👍

Good suggestions. :) I addressed them all. Thanks for taking the time.

ccoVeille:
Unless I'm wrong, you didn't apply these suggestions

https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749230173

https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749230997

https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749230384

https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749230628

jorgerojas26:
> Unless I'm wrong, you didn't apply these suggestions
> 
> [#103 (comment)](https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749230173)
> 
> [#103 (comment)](https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749230997)
> 
> [#103 (comment)](https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749230384)
> 
> [#103 (comment)](https://github.com/jorgerojas26/lazysql/pull/103#discussion_r1749230628)

You are right haha i missed them.

ccoVeille (approved):

components/Sidebar.go
```diff
@@ -0,0 +1,281 @@
...
+
+	field.SetFocusFunc(func() {
+		_, y, _, h := field.GetRect()
+		_, _, _, mph := sidebar.GetRect()
+
+		if y >= mph {
+			hidingFieldIndex := 0
+			fieldCount := sidebar.Flex.GetItemCount()
+
+			for i := 0; i < fieldCount; i++ {
```
ccoVeille, line 83: This is a candidate for `for range` loop, I think

```suggestion
			for i := range fieldCount {
```

ccoVeille (approved):
Minor remark, but OK 👍

components/Sidebar.go
jorgerojas26: I will need to upgrade go version in go.mod from 1.20 to 1.22. I don't know how much of it is a problem.

ccoVeille: You are providing a tool, not a lib.

So upgrading to 1.23 directly should be fine.

You will have to upgrade a lot of thing such as golangci-lint in the CI

While it's a thing to do, there is no need to do it in this PR, just to use `for range`

So, yes, you could upgrade, but not necessarily now

PR #120, opened 2024-11-12
Add connecting to database by connectionString given as an argument to cli

Kuchteq:
Thanks for the amazing tool! 
I'd like to see a feature like this get merged as I find it easier to just pass the path to the database I want to open as an argument, so I made this request. This implementation works but perhaps is doing a bit too much in the main.go file. I first tried doing most of this connection initialization inside Pages.go but the init() functions run before main.go's main() function.

I'd love to hear your feedback on this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,10 +132,14 @@ makepkg -si
 <!-- USAGE EXAMPLES -->
 
 ## Usage
-
 ```bash
 $ lazysql
 ```
+To launch lazysql with the ability to pick from the saved connections.
+```bash
+$ lazysql [connection_url]
+```
+To launch lazysql and connect to database at [connection_url].
 
 <p align="right">(<a href="#readme-top">back to top</a>)</p>
 
--- a/README.md
+++ b/README.md
@@ -133,12 +133,12 @@ makepkg -si
 
 ## Usage
 
-```bash
+```console
 $ lazysql
 ```
 
 To launch lazysql with the ability to pick from the saved connections.
-```bash
+```console
 $ lazysql [connection_url]
 ```
 
--- a/README.md
+++ b/README.md
@@ -132,13 +132,16 @@ makepkg -si
 <!-- USAGE EXAMPLES -->
 
 ## Usage
+
 ```bash
 $ lazysql
 ```
+
 To launch lazysql with the ability to pick from the saved connections.
 ```bash
 $ lazysql [connection_url]
 ```
+
 To launch lazysql and connect to database at [connection_url].
 
 <p align="right">(<a href="#readme-top">back to top</a>)</p>
--- a/components/ArgConnection.go
+++ b/components/ArgConnection.go
@@ -1,8 +1,8 @@
 package components
 
 import (
+	"errors"
 	"fmt"
-	"os"
 	"strings"
 
 	"github.com/jorgerojas26/lazysql/drivers"
@@ -10,11 +10,10 @@ import (
 	"github.com/jorgerojas26/lazysql/models"
 )
 
-func InitFromArg(connectionString string) {
+func InitFromArg(connectionString string) error {
 	parsed, err := helpers.ParseConnectionString(connectionString)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Could not parse connection string: %s\n", err)
-		os.Exit(1)
+		return errors.New(fmt.Sprintf("Could not parse connection string: %s", err))
 	}
 	DBName := strings.Split(parsed.Normalize(",", "NULL", 0), ",")[3]
 
@@ -40,8 +39,8 @@ func InitFromArg(connectionString string) {
 	err = newDbDriver.Connect(connection.URL)
 
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Could not connect to database %s: %s\n", connectionString, err)
-		os.Exit(1)
+		return errors.New(fmt.Sprintf("Could not connect to database %s: %s", connectionString, err))
 	}
 	MainPages.AddAndSwitchToPage(connection.URL, NewHomePage(connection, newDbDriver).Flex, true)
+	return nil
 }
--- a/components/ArgConnection.go
+++ b/components/ArgConnection.go
@@ -1,7 +1,6 @@
 package components
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 
@@ -13,7 +12,7 @@ import (
 func InitFromArg(connectionString string) error {
 	parsed, err := helpers.ParseConnectionString(connectionString)
 	if err != nil {
-		return errors.New(fmt.Sprintf("Could not parse connection string: %s", err))
+		return fmt.Errorf("Could not parse connection string: %s", err)
 	}
 	DBName := strings.Split(parsed.Normalize(",", "NULL", 0), ",")[3]
 
@@ -39,7 +38,7 @@ func InitFromArg(connectionString string) error {
 	err = newDbDriver.Connect(connection.URL)
 
 	if err != nil {
-		return errors.New(fmt.Sprintf("Could not connect to database %s: %s", connectionString, err))
+		return fmt.Errorf("Could not connect to database %s: %s", connectionString, err)
 	}
 	MainPages.AddAndSwitchToPage(connection.URL, NewHomePage(connection, newDbDriver).Flex, true)
 	return nil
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -2,15 +2,17 @@ package main
...
-
 	"github.com/jorgerojas26/lazysql/app"
 	"github.com/jorgerojas26/lazysql/components"
+	"github.com/jorgerojas26/lazysql/drivers"
+	"github.com/jorgerojas26/lazysql/helpers"
 	"github.com/jorgerojas26/lazysql/helpers/logger"
+	"github.com/jorgerojas26/lazysql/models"
+	"io"
+	"log"
+	"os"
```
ccoVeille, line 15: Fix the import orders, please

jorgerojas26:
I think your intuition is right, the main.go file need to be keep small, can you please move the command args processing to another file?. Also, can you please expand the "usage" section in the readme to include this?

Thank you for the PR.

components/ArgConnection.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/jorgerojas26/lazysql/drivers"
+	"github.com/jorgerojas26/lazysql/helpers"
+	"github.com/jorgerojas26/lazysql/models"
+)
+
+func InitFromArg(connectionString string) {
+	parsed, err := helpers.ParseConnectionString(connectionString)
+	if err != nil {
+		fmt.Printf("Could not parse connection string: %s\n", err)
```
ccoVeille, line 15:
```suggestion
		fmt.Fprintf(os.Stderr, "Could not parse connection string: %s\n", err)
```

components/ArgConnection.go
```diff
@@ -0,0 +1,40 @@
...
+		newDbDriver = &drivers.MySQL{}
+	case drivers.DriverPostgres:
+		newDbDriver = &drivers.Postgres{}
+	case drivers.DriverSqlite:
+		newDbDriver = &drivers.SQLite{}
+	}
+	err = newDbDriver.Connect(connection.URL)
+
+	if err != nil {
+		fmt.Printf("Could not connect to database %s: %s\n", connectionString, err)
```
ccoVeille, line 36: Same errors should be sent to stderr

Kuchteq:
I am facing an issue where only sqlite databases work with passing the argument. Postgresql connects without any problem but can't open unfold the left menu for listing the tables.

jorgerojas26:
I have tested MySQL, Postgres and SQLite and it works fine. Is there anything you want to add or are we ready for merge?

Kuchteq:
I don't have anything more. I'd be happy to see it merged :)

components/ArgConnection.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/jorgerojas26/lazysql/helpers"
+	"github.com/jorgerojas26/lazysql/models"
+)
+
+func InitFromArg(connectionString string) {
+	parsed, err := helpers.ParseConnectionString(connectionString)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Could not parse connection string: %s\n", err)
+		os.Exit(1)
+	}
```
ccoVeille, line 18: Please update the method signature to return an error

Calls to os.Exit should only be in main.go

Otherwise, code cannot be tested/reused

main.go
```diff
@@ -42,12 +42,13 @@ func main() {
...
 	// check if "version" arg is passed
 	argsWithProg := os.Args
-
 	if len(argsWithProg) > 1 {
 		switch argsWithProg[1] {
 		case "version":
 			println("LazySQL version: ", version)
 			os.Exit(0)
+		default:
+			components.InitFromArg(argsWithProg[1])
```
ccoVeille, line 51: So, I'm talking about adding the fmt.Println(os.Stderr, err) + os.Exit(1) here

ccoVeille (changes requested):

README.md
```diff
@@ -136,6 +136,13 @@ makepkg -si
 $ lazysql
 ```
 
+To launch lazysql with the ability to pick from the saved connections.
+```bash
+$ lazysql [connection_url]
+```
```
ccoVeille, line 142: Use


````suggestion
```console
$ lazysql [connection_url]
```
````

bash or shell if for highlighting code that is shell code, not command launched in the console

File: main.go [hunk 7]
```diff
@@ -48,6 +49,12 @@ func main() {
 		case "version":
 			println("LazySQL version: ", version)
 			os.Exit(0)
+		default:
+			err := components.InitFromArg(argsWithProg[1])
```
ccoVeille, line 53: I'm doubtful about how the code react if you pass more than one argument, the vode doesn't seem logic now

Kuchteq: I'm not sure what should be done in this scenario, the user will connect to the first database passed there and ignore the rest. Should I just then make a check if number of arguments is exactly two and display an error if it's too many?

ccoVeille: I think you should report an error if there are more than 2 arguments.

Also use could do strange things such `ladygit version somealias`

Kuchteq: I would argue that the version "subcommand" could be turned into a -v or -version flag. I don't think there is a standardized way that golang clis follow when it comes to things like these. For [lazygit](https://github.com/jesseduffield/lazygit), version is displayed via the -v or --version command, [lf](https://github.com/gokcehan/lf/tree/master) has it as -version flag. I'd say most of the cli's in general have it as the -v flag and maybe we should make the change for lazysql as well. Extra factor to take into account is that now with the patch is that if someone were to have a sqlite database called _version_ and wanted to connect to it directly, it wouldn't work as expected (realistically not the gravest issue but hey).

README.md
Kuchteq: Thank you, that clears up a lot!

main.go
ccoVeille: I agree with you.

But no matter how the version is handled, the code should handle multiple argument as an error

Kuchteq: I just mentioned it as it changes the way we parse positional arguments. Made the appropriate changes so that it errors out on multiple connection urls.

main.go
```diff
@@ -39,14 +56,16 @@ func main() {
...
+	if len(args) > 1 {
+		fmt.Fprintln(os.Stderr, "Only a single connection is allowed")
+		os.Exit(1)
+	} else if len(args) == 1 {
+		err := components.InitFromArg(args[0])
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
 		}
 	}
```
ccoVeille, line 70: I would like to suggest something like this.

```suggestion
	switch len(args) {
		case 0:
			// nothing to do
    case 1:
			err := components.InitFromArg(args[0])
			if err != nil {
				fmt.Fprintln(os.Stderr, err)
				os.Exit(1)
 			}
		default:
			fmt.Fprintln(os.Stderr, "Only a single connection is allowed")
			os.Exit(1)
 	}
```



The reason is that testing the same variable in multiple if statement makes it a perfect candidate to use a switch

main.go
```diff
@@ -39,14 +56,16 @@ func main() {
...
-			fmt.Println("LazySQL version: ", version)
-			os.Exit(0)
+	if len(args) > 1 {
+		fmt.Fprintln(os.Stderr, "Only a single connection is allowed")
+		os.Exit(1)
+	} else if len(args) == 1 {
+		err := components.InitFromArg(args[0])
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
```
ccoVeille, line 68: log.Fatal is already used on errors, so you should reuse it

```suggestion
			log.Fatal(err)
```

log.Fatal print errors to stderr too, then call os.Exit(1)

main.go
```diff
@@ -39,14 +56,16 @@ func main() {
...
+	args := flag.Args()
 
-	if len(argsWithProg) > 1 {
-		switch argsWithProg[1] {
-		case "version":
-			fmt.Println("LazySQL version: ", version)
-			os.Exit(0)
+	if len(args) > 1 {
+		fmt.Fprintln(os.Stderr, "Only a single connection is allowed")
+		os.Exit(1)
```
ccoVeille, line 63: See https://github.com/jorgerojas26/lazysql/pull/120#discussion_r1871984775

main.go
Kuchteq: Good practice! Updated and I'll keep it in mind for future changes :))

jorgerojas26:
Thank you @ccoVeille for the good advices. I'm learning too, so thanks for taking the time to keep an eye on the project.

ccoVeille (approved):

PR #123, opened 2024-11-16
feat: Refresh tree shortcut

### LINKED ISSUE #118: Creating a table does not reflect in the TUI
When you use the `SQL Editor` to create a table, it does not reflect inside the TUI and you have to restart lazysql to see the changes.

# Reproduction

> This was tested with SQLite

1. open `lazysql`
2. connect to a database
3. open SQL Editor and execute `CREATE TABLE` query
    - `CREATE TABLE test (id integer PRIMARY KEY);`
 4. switch back to table list -> new table is not visible

# Expectation
It would be nice to 'refresh' the table list after a new table was created, and maybe even add a shortcut for manual refreshing of the table list.

Would it be possible to filter the query-string for 'CREATE' or 'create' and if either one of them is present refresh the table list after the query was executed?

gytic:
This PR should fix #118 

To refresh the tree view my idea is to:
1. clear existing nodes (under root)
2. Re-initialize nodes (under root)
3. Reset search state
4. Close opened table if it no longer exists

### DIFF
--- a/components/Tree.go
+++ b/components/Tree.go
@@ -547,46 +547,29 @@ func (tree *Tree) InitializeNodes(dbName string) {
 		databases = []string{dbName}
 	}
 
-	if tree.GetSelectedDatabase() == "" {
-		for _, database := range databases {
-			childNode := tview.NewTreeNode(database)
-			childNode.SetExpanded(false)
-			childNode.SetReference(database)
-			childNode.SetColor(app.Styles.PrimaryTextColor)
-			rootNode.AddChild(childNode)
+	for _, database := range databases {
+		childNode := tview.NewTreeNode(database)
+		childNode.SetExpanded(false)
+		childNode.SetReference(database)
+		childNode.SetColor(app.Styles.PrimaryTextColor)
+		rootNode.AddChild(childNode)
 
-			go func(database string, node *tview.TreeNode) {
-				tables, err := tree.DBDriver.GetTables(database)
-				if err != nil {
-					logger.Error(err.Error(), nil)
-					return
-				}
+		go func(database string, node *tview.TreeNode) {
+			tables, err := tree.DBDriver.GetTables(database)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


gytic:
@jorgerojas26 do you have an idea where and how to close the opened table if it was deleted?

I will look tomorrow a bit more into the code base to figure out how it is opened etc but maybe you have an idea already how one could implement that.

jorgerojas26:
I'm not sure what you mean but maybe you just delete the tab that has the table open.

gytic:
Okay I think this is the best I can do.

When refreshing, all nodes under the root node get deleted and afterwards re added.

components/Tree.go
```diff
@@ -557,3 +528,48 @@ func (tree *Tree) ExpandAll() {
...
+			tree.databasesToNodes(tables, node, true)
+			App.Draw()
+		}(database, childNode)
+	}
+}
+
+func (tree *Tree) Refresh(dbName string) {
+	rootNode := tree.GetRoot()
+	rootNode.ClearChildren()
+	// readd nodes
```
ccoVeille, line 573:
```suggestion
	// read nodes
```

Or


```suggestion
	// re-add nodes
```

Using readd is ambiguous

gytic: this is fixed now, thanks for pointing out

gytic:
PR is ready for review

jorgerojas26:
This is good enough for now, i think. Can you please just change the keybinding from 'r' to 'R'. The reason is, there is a 'R' keymap for the table that refreshes the data, so, to keep consistency let's use 'R' also for the tree.

With that change, i'm ready to merge.

Thank you very much

gytic:
I will work on it this evening

gytic:
I ran the lints locally and had no errors. I think this should be now fixed.

Also changed the shortcut from `r` to `R`.
PR should be ready for merge now.

PR #124, opened 2024-11-17
FEAT: Improve behavior of search `/` function on tree

### LINKED ISSUE #121: [FEAT] Search tables (`/`) within a specific database
Currently when searching tables using `/`, It will search for the table throughout all of the databases. 

I'm working with a server with multiple databases that has a same table names, I find it hard to search for the table in the database that I'm looking for. 

It would be nice if we can search tables within a specific database. Or just be able to have a view on which database is this table from when searching (if there's a lot of tables on the database, you can't see the database name if the table is on the bottom part)

KevsterAmp:
- Closes #121

# Current Behavior
Running the search command `/` will expand all databases and find all possibilities of tables within the whole server, throughout all databases.

The issue arises when there are multple instances of table names within multiple databases. it becomes hard to determine which database the focused table is from since the database is expanded, and if there are 20+ tables you cannot see the DB name unless you move up. 

# Implemented Behavior
To mark a database as "selected" It has to be the `currNode` and it should be expanded. 

- If there is no database selected. The search function will search throughout the names of all the databases ONLY.
- If there is a database selected. The search function will search thorughout the names of all the tables within the selected database.

I believe this is a better implementation since its most common to search a table within a specific database, rather than all the databases. 

I can also include in this PR to move the command that "searches on all databases" to `CTRL + /`.

### DIFF
--- a/components/Tree.go
+++ b/components/Tree.go
@@ -326,28 +326,33 @@ func (tree *Tree) databasesToNodes(children map[string][]string, node *tview.Tre
 
 func (tree *Tree) search(searchText string) {
 	rootNode := tree.GetRoot()
+	currNode := tree.GetCurrentNode()
 	lowerSearchText := strings.ToLower(searchText)
 	tree.state.searchFoundNodes = []*tview.TreeNode{}
 
 	if lowerSearchText == "" {
 		rootNode.Walk(func(_, parent *tview.TreeNode) bool {
 			if parent != nil && parent != rootNode && parent.IsExpanded() {
-				parent.SetExpanded(false)
+				parent.SetExpanded(true)
 			}
 			return true
 		})
 		return
 	}
 
-	// filteredNodes := make([]*TreeStateNode, 0, len(treeNodes))
+	var searchStartNode *tview.TreeNode
+	if currNode != nil && currNode.IsExpanded() {
+		// search between tables if theres a curentdatabase selected
+		searchStartNode = currNode
+	} else {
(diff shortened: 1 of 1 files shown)


KevsterAmp:
Found a bug that I'm not sure how to fix.
When you expand some database nodes, the app seems to record all the tables within it. So when you run a search command even if there are no selected database. It can show results on the tables within the db itself. 

At first glance, i think has to be related `tree.state.searchFoundNodes`. I'll try to look at the issue more later in the day

KevsterAmp:
Thoughts on this? @jorgerojas26 Thanks

components/Tree.go
```diff
@@ -339,15 +340,19 @@ func (tree *Tree) search(searchText string) {
...
 
-	// filteredNodes := make([]*TreeStateNode, 0, len(treeNodes))
+	var searchStartNode *tview.TreeNode
+	if currNode != nil && currNode.IsExpanded() {
+		// search between tables if theres a curentdatabase selected
+		searchStartNode = currNode
+	} else {
+		// if there's no current database selected, search within the database names only
+		searchStartNode = rootNode
+	}
```
ccoVeille, line 350:
```suggestion
	// search within the database names only by default
	searchStartNode := rootNode
	if currNode != nil && currNode.IsExpanded() {
		// search between tables if there is a database selected
		searchStartNode = currNode
	}
```

jorgerojas26:
Closing in favor of #165

PR #126, opened 2024-11-20
Fix query to get primary column names for postgres

svanharmelen:
Not sure if/how this worked for other users, but for my setup (using named schemas) this didn't work. By taking the schema name into account when doing the query it now works great for all use cases (as far as I could test/verify).

### DIFF
--- a/drivers/postgres.go
+++ b/drivers/postgres.go
@@ -861,9 +861,12 @@ func (db *Postgres) GetPrimaryKeyColumnNames(database, table string) (primaryKey
 
 	splitTableString := strings.Split(table, ".")
 
-	if len(splitTableString) == 1 {
+	if len(splitTableString) != 2 {
 		return nil, errors.New("table must be in the format schema.table")
 	}
+
+	schemaName := splitTableString[0]
+	tableName := splitTableString[1]
 
 	if database != db.CurrentDatabase {
 		err = db.SwitchDatabase(database)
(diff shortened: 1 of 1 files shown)


drivers/postgres.go
```diff
@@ -863,32 +863,40 @@ func (db *Postgres) GetPrimaryKeyColumnNames(database, table string) (primaryKey
 
 	splitTableString := strings.Split(table, ".")
 
-	if len(splitTableString) == 1 {
+	if len(splitTableString) != 2 {
```
svanharmelen, line 866: This seems like a more robust way to determine of a string in the format of `schema.table` was given.

drivers/postgres.go
```diff
@@ -863,32 +863,40 @@ func (db *Postgres) GetPrimaryKeyColumnNames(database, table string) (primaryKey
...
-			_ = db.SwitchDatabase(db.PreviousDatabase)
-		}
-	}()
 
-	tableName := splitTableString[1]
+		defer func() {
+			if r := recover(); r != nil {
+				_ = db.SwitchDatabase(db.PreviousDatabase)
+			}
+		}()
```
svanharmelen, line 883: Not sure why you would want to catch a panic here? Or did you intent to call the `db.SwitchDatabase(db.PreviousDatabase)` method in case this method returned an error? Or exactly the other way around maybe (only call the `db.SwitchDatabase(db.PreviousDatabase)` method when this method returned without an error 😅)?

Either was I would expect that switching back would only be needed is we actually switched in the first place 😏

jorgerojas26: I admit i find that code a bit weird, i wanted to switch back to the previous database in case of any error, but maybe there's a better way.

svanharmelen: Ah check... In that case I will make a PR with a suggested way to achieve that goal in a more idiomatic approach. I will remove this change from this PR so you can look at them in isolation.

PR #128, opened 2024-11-21
Add ILIKE to the list of comparators

svanharmelen:


components/ResultsTable.go
```diff
@@ -1220,7 +1220,7 @@ func (table *ResultsTable) search() {
 
 	table.Filter.Input.SetAutocompleteFunc(func(currentText string) []string {
 		split := strings.Split(currentText, " ")
-		comparators := []string{"=", "!=", ">", "<", ">=", "<=", "LIKE", "NOT LIKE", "IN", "NOT IN", "IS", "IS NOT", "BETWEEN", "NOT BETWEEN"}
+		comparators := []string{"=", "!=", ">", "<", ">=", "<=", "LIKE", "ILIKE", "NOT LIKE", "IN", "NOT IN", "IS", "IS NOT", "BETWEEN", "NOT BETWEEN"}
```
ccoVeille, line 1223: I'm surprised do not see REGEXP and NOT REGEXP

ccoVeille, line 1223: Why not?


```suggestion
		comparators := []string{"=", "!=", ">", "<", ">=", "<=", "LIKE", "NOT LIKE", "ILIKE", "NOT ILIKE", "IN", "NOT IN", "IS", "IS NOT", "BETWEEN", "NOT BETWEEN"}
```

svanharmelen: Sure, why not 😉

svanharmelen: Added them as well while I was at it...

svanharmelen:
I just added a few more missing components and improved the completion a little so it actually filters out suggestions that no longer match what a user has typed in. In addition ordered the suggestions for readability.

Will make a separate PR for fixing the lint error. Guess the linter is updated as this PR doesn't touch that code.

PR #129, opened 2024-11-21
Add ability to run commands before trying to connect

svanharmelen:
While the code is a bit "quick and dirty", it works like a charm and allows you to define connections like this:

```toml
[[database]]
Name = 'server'
Provider = 'postgres'
DBName = 'foo'
URL = 'postgres://postgres:password@localhost:${port}/foo'
Commands = [
  'ssh remote-bastion -L ${port}:localhost:5432'
]
```

So you can easily connect to remote servers as well! You can even chain commands to connect to a remote server and then a postgres container running in a remote k8s cluster:

```toml
[[database]]
Name = 'container'
Provider = 'postgres'
DBName = 'foo'
URL = 'postgres://postgres:password@localhost:${port}/foo'
Commands = [
  'ssh remote-bastion -L 6443:localhost:6443', # first forward the kube API port 
  'kubectl port-forward service/postgres ${port}:5432 --kubeconfig /path/to/kube.conf'
]
```


components/ConnectionSelection.go
```diff
@@ -140,50 +143,69 @@ func NewConnectionSelection(connectionForm *ConnectionForm, connectionPages *mod
...
+	if len(connection.Commands) > 0 {
+		cs.StatusText.SetText("Running commands...").SetTextColor(app.Styles.TertiaryTextColor)
 		App.Draw()
 
-		var newDbDriver drivers.Driver
+		// This is super arbitrary, but allows you to use '${port}' in your commands
+		// which will be replaced with a random port number between 5000 and 6000 to
+		// avoid conflicts when connecting to multiple databases from the same machine.
+		port := 5000 + rand.Intn(1000)
+		connection.URL = strings.ReplaceAll(connection.URL, "${port}", fmt.Sprintf("%d", port))
```
ccoVeille, line 162: This is a bit dirty and not fail proof

Try something like this

https://gist.github.com/sevkin/96bdae9274465b2d09191384f86ef39d

Thanks @sekvin BTW

svanharmelen: Guess this indeed is a little less dirty 😉 Added an updated version of the example and used that to acquire a port.

svanharmelen:
I've noticed this needs a little bit more work to make it robust (does not cleanly stops all commands when making multiple connection in one session). Will make some tweaks this weekend...

svanharmelen:
OK... It's quite a bit more involved as before as I needed some plumbing to be able to keep track of running commands in such a way that they would cleanup before the program exited. With these changes it now works very nicely and no matter how the app is stopped (pressing `q`, `Ctrl-C` or sending `SIGINT` or `TERM`) the commands are always properly stopped.

Next to that I added some logic for testing if a command started listening on the requested port (see the updated example in the opening post of this PR) and tried to provide a little bit more feedback to the user.

Besides the functional changes needed for the commands to work properly, I think I make 3 semi unrelated changes. If you want I can revert them, but I think they make sense:

1. I removed the `panic`'s from `main.go`
2. I renamed `loglvl` to `loglevel` so its more consistent with `logfile` (which is also written in full)
3. I updated the label on `home.Tree` from `connection.Name`+`connection.Provider` to just `connection.Name`.

The last one is mainly because practically all my names are too long for the tree width and there is a bug in how the cutoff is handled so next to loosing the last part (which makes sense) I also don't see the first few chars. I would ask/suggest to just use the `connection.Name` for now and maybe revert that once the behavior is fixed or changed.

Let me know if you have any questions or comments.

ccoVeille:
The code is way better now thanks.

I might have some remarks on the way you use context. But there are side remarks.

It could be addressed in a later refactoring.

Here are my feedbacks:
- the context shouldn't be stored in Application, but passed from one method to another.
- I'm unsure you need to stored the context canceller in the struct either.
- you could move the signal check things in a method that create the context.
- The context could be created in main.go
- waitforport should pass the context. Avoiding to retry if the context is done.

helpers/utils.go
```diff
@@ -18,3 +23,32 @@ func ContainsCommand(commands []commands.Command, command commands.Command) bool
...
+	for i := 0; i < 10; i++ {
+		conn, err := net.DialTimeout("tcp", "localhost:"+port, 500*time.Millisecond)
+		if err == nil {
+			_ = conn.Close()
+			return nil
+		}
+		time.Sleep(500 * time.Second)
+	}
+	return errors.New("Timeout waiting for port " + port)
+}
```
ccoVeille, line 54:
```suggestion
func WaitForPort(ctx context.Context, port string) error {
	for i := 0; i < 10; i++ {
		conn, err := net.DialTimeout("tcp", "localhost:"+port, 500*time.Millisecond)
		if err == nil {
			_ = conn.Close()
			return nil
		}

		select {
			case <- time.After(500 * time.Second):
			case <- ctx.Done():
				return ctx.Err()
		}
	}
	return errors.New("Timeout waiting for port " + port)
}
```

svanharmelen:
I'm well aware of how the context could/should be used in a well designed application, but properly implementing it into the current codebase would be close to a full rewrite 🙈

So I opted to not try and change everything at once, but instead focus on the bits I wanted to add and make small improvements in the surrounding parts to make that possible.

ccoVeille:
I agree. That's why I said your change were OK, and the code is way better now.

And yes, it might need a pure rewrite

helpers/utils.go
svanharmelen: While I doubt if this can ever have a blocking effect on the app (this is isn't called from the main thread) I agree its the more "correct" way to do this.

Even more correct would be to pass the context to the dialer as well 😏 But again, this code will never be blocking the app so doubt if it adds any value.

But I can have a quick look tomorrow and add that last bit.

svanharmelen:
Updated `WaitForPort` so it takes a context which is set on the dailer and is checked when sleeping in between connection attempts. There are still many things that can be added and/or improved regarding both the code and functionality in this PR, but for now I think this provides enough functionality for most use cases.

helpers/utils.go
```diff
@@ -18,3 +24,43 @@ func ContainsCommand(commands []commands.Command, command commands.Command) bool
...
+	for i := 0; i < 10; i++ {
+		conn, err := dialer.DialContext(ctx, "tcp", "localhost:"+port)
+		if err == nil {
+			_ = conn.Close()
+			return nil
+		}
+
+		select {
+		case <-time.After(500 * time.Millisecond):
+			continue
```
svanharmelen, line 59: I know this is not needed here, but I like the expressiveness of it.

ccoVeille (approved):

ccoVeille:
> I'm well aware of how the context could/should be used in a well designed application, but properly implementing it into the current codebase would be close to a full rewrite 🙈
> 
> So I opted to not try and change everything at once, but instead focus on the bits I wanted to add and make small improvements in the surrounding parts to make that possible.
> 
> I actually started with a function that returns a context in main, more of less exactly how you are proposing 😉, but the current layout of the app is not realy designed to work that way as it uses lots of global variables to store types (like the main Application type) that are being called from locations all over the code.
> 
> Properly restructering that in a more idiomatic layout would be a reasonably big change that should be a PR on itself in my opinion (and should be endorced by @jorgerojas26 as the owner/maintainer of this code).
> 
> _Edit: Not proposing such a rewrite by the way! I like the tool and love to fix bugs that bug me or contribute features I'm missing, but I do have a (quite) busy day job as well 😊_

Maybe the discussions we had here, could be quoted in a dedicated issue about a possible refactoring that would make the app more idiomatic, especially the context.

The issue might be closed then,but at least there will be a track of the discussion in an issue

jorgerojas26:
> > I'm well aware of how the context could/should be used in a well designed application, but properly implementing it into the current codebase would be close to a full rewrite 🙈
> > So I opted to not try and change everything at once, but instead focus on the bits I wanted to add and make small improvements in the surrounding parts to make that possible.
> > I actually started with a function that returns a context in main, more of less exactly how you are proposing 😉, but the current layout of the app is not realy designed to work that way as it uses lots of global variables to store types (like the main Application type) that are being called from locations all over the code.
> > Properly restructering that in a more idiomatic layout would be a reasonably big change that should be a PR on itself in my opinion (and should be endorced by @jorgerojas26 as the owner/maintainer of this code).
> > _Edit: Not proposing such a rewrite by the way! I like the tool and love to fix bugs that bug me or contribute features I'm missing, but I do have a (quite) busy day job as well 😊_
> 
> Maybe the discussions we had here, could be quoted in a dedicated issue about a possible refactoring that would make the app more idiomatic, especially the context.
> 
> The issue might be closed then,but at least there will be a track of the discussion in an issue

That is a really good idea. I sadly did not know about Context api when i started building lazysql (remember this is my first touch on Go). So, that refactor would be a good idea.

svanharmelen:
@jorgerojas26 just added a commit with some documentation, so guess this one should be ready now as well.

PR #130, opened 2024-11-21
Make the items in the tree alphabetically ordered

svanharmelen:


components/Tree.go
```diff
@@ -291,7 +292,20 @@ func NewTree(dbName string, dbdriver drivers.Driver) *Tree {
 func (tree *Tree) databasesToNodes(children map[string][]string, node *tview.TreeNode, defaultExpanded bool) {
 	node.ClearChildren()
 
-	for key, values := range children {
+	// Sort the keys and use them to loop over the
+	// children so they are always in the same order.
+	sortedKeys := make([]string, 0, len(children))
+	for key := range children {
+		sortedKeys = append(sortedKeys, key)
+	}
```
ccoVeille, line 300: That's [maps.Keys](https://pkg.go.dev/maps#Keys) (go 1.23)


```suggestion
	sortedKeys := maps.Keys(children)
```

ccoVeille (approved):

components/Tree.go
svanharmelen: Well in that case it should be this I guess: `sortedKeys := slices.Sorted(maps.Keys(children))`

As this is a binary and not I lib, I guess there is no harm in updating the Go version. I'll update the PR and will see what @jorgerojas26 thinks about it.

jorgerojas26: Yes, i'm fine with updating Go version.

PR #131, opened 2024-11-21
Fix a copy/base bug in the query editor

svanharmelen:
Next to the fact that pasting to the query editor was extremely slow, it also filtered out some chars like `{` and `}` which broke queries pasted in. These things where actually already updated/fix in the used dependencies, so updating them fixed these issues.


jorgerojas26:
Please remove the version update of tview package and we are ready to merge

svanharmelen:
May I ask why you don't want to update that one? As that is actually the main package that needed to be updated in order to solve the issue 😊

jorgerojas26:
You are right, i'm dumb haha.

PR #133, opened 2024-11-22
Do not try to recover from panics, but use the return values

svanharmelen:
Sorry for all the PRs, I'm just having fun with it and am hitting a lot of small issues when using `lazysql` which I would like to help improve 😉 

This PR removes all the calls to `recover` as it is very discouraged (see Go docs) to try and recover panics as in in 99.999999% of the cases that indicates a bug in the program or another error not being handled properly which should be fixed instead.

In order to take any actions using a `defer` that will only do something if the function returned an error, its best to just check against that error. While in Go its no too common to use named return values in the function signature, in the case where you want a defer to have access to the actual returned value I found it to be the most robust to at least name the error in the function signature. But in this case I noticed you already named _all_ return values so I didn't need to change anything for that.

### DIFF
--- a/components/ResultsTable.go
+++ b/components/ResultsTable.go
@@ -812,23 +812,11 @@ func (table *ResultsTable) SetResultsInfo(text string) {
 }
 
 func (table *ResultsTable) SetLoading(show bool) {
-	defer func() {
-		if r := recover(); r != nil {
-			logger.Error("ResultsTable.go:800 => Recovered from panic", map[string]any{"error": r})
-			_ = table.Page.HidePage(pageNameTableLoading)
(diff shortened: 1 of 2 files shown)


jorgerojas26:
That is fair enough :) just fix the linter issues and i'll merge.

jorgerojas26:
Thanks for all the PRs, by the way. I'm just a bit buse right now, but i'll try to follow all of them.

svanharmelen:
The linter issue is fixed by #132 I think the linter updated as I didn't tough that code...

svanharmelen:
Addressed all feedback and rebased all PRs so they pass the linter again 👍🏻

PR #134, opened 2024-11-30
Add a global tree search keyboard shortcut

itsmewes:
Hi,

I've been really loving lazysql and use it as my daily driver now. Thank you for the time you've put into it!

One thing that I've been missing is to be able to search the tables from anywhere. Right now if I want to go to another table from the results table, I have to press, `<shift>+h` then `/` then `<ctrl>+u` to search for a new table, which for me is a bit of a chore to do. In TablePlus, you can hit `<cmd>+p` (on a macOS) to search for any table, though I have used `<ctrl+p>` since that seems to be the modifier key of choise. I have (probably very crudely) added that here. My Go and knowledge of the code-base is not great, so thinking of this PR as an initial discussion on this suggested key binding.

Please let me know what you think and if you are ok with this kind of keybinding being added how to best go about it.


jorgerojas26:
Thanks for the PR. I'll take a look.

PR #142, opened 2024-12-13
Introduce a way to support user defined configurations

svanharmelen:
This is by no means perfect, but its again taking a few small steps to make minor improvements and introduce some new functionality which I was really missing.


svanharmelen:
Do you need anything from me in order to get this one merged?

jorgerojas26:
Sorry for the delay (christmas). I'll try to review it as soon as i can.

app/config.go
```diff
@@ -10,38 +10,33 @@ import (
...
-
-	err = toml.Unmarshal(file, &config)
-
-	return
 }
 
-func LoadConnections() (connections []models.Connection, err error) {
-	config, err := LoadConfig()
+func LoadConfig() error {
+	file, err := os.ReadFile(filepath.Join(os.Getenv("HOME"), ".config", "lazysql", "config.toml"))
```
ccoVeille, line 26: Please take a look at os.UserConfigDir

https://pkg.go.dev/os#UserConfigDir

Look at the code
https://cs.opensource.google/go/go/+/go1.23.4:src/os/file.go;l=528

The implementation is there for all OSs

app/config.go
```diff
@@ -10,38 +10,33 @@ import (
...
-	return
+	return toml.Unmarshal(file, App.config)
 }
 
-func SaveConnectionConfig(connections []models.Connection) (err error) {
-	config := Config{Connections: connections}
+func (c *Config) SaveConnections(connections []models.Connection) error {
+	c.Connections = connections
 
 	directoriesPath := filepath.Join(os.Getenv("HOME"), ".config", "lazysql")
```
ccoVeille, line 36: Please consider updating this one too

app/config.go
svanharmelen: Hmm... Yeah the only issue I have with this one is that on Mac its pointing to `$HOME/Library/Application Support` instead of the more common `$HOME/.config` used for most packages you install through brew (the main packaging system for Mac). On the other hand returning `%AppData%` for Windows users is of course much, much better then `%HOME/.config`, so I'll update the code to use this instead 👍🏻

svanharmelen:
Updated as requested. Please note that depending on your OS, this is a breaking change and you will need to move your existing config file to the (potentially) new location.

app/config.go
svanharmelen: Of course...

app/config.go
```diff
@@ -0,0 +1,58 @@
...
+		},
+	}
+}
+
+func LoadConfig() error {
+	configDir, err := os.UserConfigDir()
+	if err != nil {
+		return err
+	}
+	file, err := os.ReadFile(filepath.Join(configDir, "lazysql", "config.toml"))
```
ccoVeille, line 30: You should have an unexported method to compute the config file path instead of computing it twice

app/config.go
ccoVeille: Yeah, I know.

One solution that has been implemented many times is that you could check for the XDG_CONFIG_HOME env then falling back on os.UserConfigDir when not present

app/config.go
svanharmelen: I guess you mean:

> Would probably make sense to add an unexported method to compute the config file path instead of computing it twice?

Sure why not 😉

ccoVeille (approved):

jorgerojas26:
I really don't like the breaking change. On a fresh install on MacOS, for example, the app does not work out of the box because it can't find the config file. Can we just have default values in case the file is not present?.

Also, i don't like that on macOS the config must be located in "$HOME/Library/Application Support". I'd like to keep supporting "$HOME/.config", can we search on both paths (with $HOME/.config as primary source) or maybe check for $XDG_CONFIG_HOME first?

svanharmelen:
>I really don't like the breaking change. On a fresh install on MacOS, for example, the app does not work out of the box because it can't find the config file. Can we just have default values in case the file is not present?.

Good catch! That is not how I intended it to work, of course it should work without having to first create a config file yourself, so let me update that.

>Also, i don't like that on macOS the config must be located in "$HOME/Library/Application Support". I'd like to keep supporting "$HOME/.config", can we search on both paths (with $HOME/.config as primary source) or maybe check for $XDG_CONFIG_HOME first?

OK so what do you suggest? Personally I think adding support for `$XDG_CONFIG_HOME` makes sense as it seems to be a much used way to set your preferred config dir. Making the app use `$HOME/.config` if that dir exists is also possible, but feels a bit less logical/expected. In that case I would almost argue to revert this part of the change all together or at least for macOS.

So to prevent multiple additional iterations, please let me know your preference.

jorgerojas26:
> > I really don't like the breaking change. On a fresh install on MacOS, for example, the app does not work out of the box because it can't find the config file. Can we just have default values in case the file is not present?.
> 
> Good catch! That is not how I intended it to work, of course it should work without having to first create a config file yourself (it indeed uses default values if the config file doesn't exist), so let me update that.
> 
> > Also, i don't like that on macOS the config must be located in "$HOME/Library/Application Support". I'd like to keep supporting "$HOME/.config", can we search on both paths (with $HOME/.config as primary source) or maybe check for $XDG_CONFIG_HOME first?
> 
> OK so what do you suggest? Personally I think adding support for `$XDG_CONFIG_HOME` makes sense as it seems to be a much used way to set your preferred config dir. Making the app use `$HOME/.config` if that dir exists is also possible, but feels a bit less logical/expected. In that case I would almost argue to revert this part of the change all together or at least for macOS.
> 
> So to prevent multiple additional iterations, please let me know your preference.

Thanks for the fix.

Yeah, i think supporting `$XDG_CONFIG_HOME` is the way.

svanharmelen:
Updated to code to _always_ check for a `$XDG_CONFIG_HOME` env var (so independent of the OS) and if that isn't set, then we fallback to the `os.UserConfigDir` function to get the config dir.

jorgerojas26:
Thank you, please fix the merge conflict and we are good to go.

svanharmelen:
Converting to a draft for just a minute to verify and double check if everything works as expected. Should only take a few minutes...

svanharmelen:
Did find a minor bug in the pagination logic, but will make a separate PR for that. The code in this PR now seems to work as expected.

ccoVeille:
> Updated to code to _always_ check for a `$XDG_CONFIG_HOME` env var (so independent of the OS) and if that isn't set, then we fallback to the `os.UserConfigDir` function to get the config dir.

This remind me a suggestion made here

https://github.com/jorgerojas26/lazysql/pull/142#discussion_r1899446960

😁😇🤣😂

PR #146, opened 2024-12-18
feat: support SHOW query results and Y to copy row

bobchandler:
eg: SHOW VARIABLES LIKE 'innodb_buffer_pool_size';  

Variable_name	Value
innodb_buffer_pool_size	1073741824


jorgerojas26:
Hello, thanks for the PR.

While i like the idea of the "copy as" feature, I think those are too many shortcuts to remember just for copying. I was thinking a better way to do it could be 'y' for copying the cell as text, 'Y' will copy the entire row as text (or maybe CSV) by default, and 'CTRL + y' would open a popup with a list of "copy as" options. What do you think?

bobchandler:
indeed, that is correct thinking to reduce shortcuts for copying.

bobchandler:
6d1da1fbd844d7e421d384301f35c2066f31f4b4: support show command, if not that feature, execute show command, 0 rows affected.

bobchandler:
@jorgerojas26 that is all feature

jorgerojas26:
The 'copy as (sql)' doesn't take into account the DB provider. In fact, the extraction and formatting of the query should be a Driver responsibility.

For example, Postgres does not support backticks for column identifiers, so the copied query will not work.

Maybe the Driver will need a function that receive some parameters to construct the statement, something like this:

```
func (db *Driver) buildQuery(statement string, cols string[], values string[], primaryKeyNames string[], primaryKeyValues string[]) {
    switch(statement){
        case "INSERT":
        ...
    }
}
```

But this is only one idea, i'm not at home right now (christmas) to do a full review, but i think you can get the idea.

jorgerojas26:
Hi @bobchandler, any update on this? i really like this feature.

PR #151, opened 2025-01-05
Update the pagination logic for queries done from the SQL editor

svanharmelen:
They previously didn't take the row with column names into account and I noticed there are actually 3 slightly different methods that calculate the pagination info, while I previously just updated/fixed one of them

The second commit changes the way how the SQL editor behaves a little as I continuously found myself doing the wrong thing here because the behavior was different depending on the result. If there were results the focus would switch to the table, if not it would stay with the editor.

While I understand that in some flows that makes sense as you might want to update your query and try again, but in a lot of other flows you might want to return to the tree or open another DB or...

So I thought it would easier for peoples' muscle memory to learn one consistent flow instead of having to think in between if there was output and if I should do different. For me the flow `Ctrl-E -> write or update query -> Ctrl-R` is really easy, consistent and never gives unexpected behavior.

But all that being said, let me know if you want me to drop the last commit. Thanks!

### DIFF
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -17,9 +17,12 @@ type Driver interface {
 	UpdateRecord(database, table, column, value, primaryKeyColumnName, primaryKeyValue string) error
 	DeleteRecord(database, table string, primaryKeyColumnName, primaryKeyValue string) error
 	ExecuteDMLStatement(query string) (string, error)
-	ExecuteQuery(query string) ([][]string, error)
+	ExecuteQuery(query string) ([][]string, int, error)
 	ExecutePendingChanges(changes []models.DBDMLChange) error
-	SetProvider(provider string) // NOTE: This is used to get the primary key from the database table until i find a better way to do it. See ResultsTable.go GetPrimaryKeyValue function
 	GetProvider() string
 	GetPrimaryKeyColumnNames(database, table string) ([]string, error)
+
+	// NOTE: This is used to get the primary key from the database table until I
+	// find a better way to do it. See *ResultsTable.GetPrimaryKeyValue()
+	SetProvider(provider string)
 }
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -190,7 +190,7 @@ func (db *MSSQL) GetForeignKeys(database, table string) ([][]string, error) {
 
 func (db *MSSQL) GetIndexes(database, table string) ([][]string, error) {
 	query := `
-		SELECT 
+		SELECT
 				t.name AS table_name,
 				i.name AS index_name,
 				i.is_unique AS is_unique,
@@ -201,33 +201,33 @@ func (db *MSSQL) GetIndexes(database, table string) ([][]string, error) {
 				ic.is_included_column AS is_included,
 				i.has_filter AS has_filter,
 				i.filter_definition AS filter_definition
-		FROM 
-				sys.tables t 
+		FROM
+				sys.tables t
 		INNER JOIN
-				sys.schemas s 
+				sys.schemas s
 					ON
 						t.schema_id = s.schema_id
 		INNER JOIN
-				sys.indexes i 
+				sys.indexes i
 					ON
 						t.object_id = i.object_id
 		INNER JOIN
-				sys.index_columns ic 
+				sys.index_columns ic
 					ON
 						i.object_id = ic.object_id
 					AND
 						i.index_id = ic.index_id
 		INNER JOIN
-				sys.columns c 
+				sys.columns c
 					ON
 						t.object_id = c.object_id
 					AND
 						ic.column_id = c.column_id
-		WHERE 
+		WHERE
 				DB_NAME() = @p1
 		AND
 				t.name = @p2
-		ORDER BY 
+		ORDER BY
 				i.type_desc
 	`
 	return db.getTableInformations(query, database, table)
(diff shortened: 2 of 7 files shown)


drivers/driver.go
```diff
@@ -17,7 +17,7 @@ type Driver interface {
 	UpdateRecord(database, table, column, value, primaryKeyColumnName, primaryKeyValue string) error
 	DeleteRecord(database, table string, primaryKeyColumnName, primaryKeyValue string) error
 	ExecuteDMLStatement(query string) (string, error)
-	ExecuteQuery(query string) ([][]string, error)
+	ExecuteQuery(query string) ([][]string, int, error)
```
ccoVeille, line 20: I would like to suggest to add named result in return value to make it clearer


```suggestion
	ExecuteQuery(query string) (results [][]string, totalRecords int, err error)
```

drivers/mysql.go
```diff
@@ -373,19 +373,19 @@ func (db *MySQL) GetRecords(database, table, where, sort string, offset, limit i
 		return nil, 0, err
 	}
 
-	return
+	return paginatedResults, totalRecords, nil
 }
 
-func (db *MySQL) ExecuteQuery(query string) (results [][]string, err error) {
+func (db *MySQL) ExecuteQuery(query string) (results [][]string, records int, err error) {
```
ccoVeille, line 379: This name is used in previous method, I find it more meaningful

```suggestion
func (db *MySQL) ExecuteQuery(query string) (results [][]string, totalRecords int, err error) {
```

drivers/mssqql.go
```diff
@@ -435,10 +435,10 @@ func (db *MSSQL) ExecuteQuery(query string) ([][]string, error) {
 	}
 
 	if err := rows.Err(); err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 
-	return results, nil
+	return results, len(results) - 1, nil
```
ccoVeille, line 441: I don't get the -1 (here and everywhere)

If you want to return the number of records, then it's len(results) only

If you do a -1, then it's an offset.

I'm unsure what are trying to do/get

Maybe I'm simply lost and my comment doesn't make sense

Maybe what you want to do (so something around the -1) has to be done in the code calling these methods and not here

drivers/driver.go
svanharmelen: Well... No other methods in this interface use named return values so it sounds like that suggested improvement is not really related to my changes (and introducing it just for this method would make the code inconsistent which is not something I'd like to do).

So instead, if you think its a valuable addition, I would suggest that you open a PR yourself to update all methods in this interface in one go so everything will remain consistent and the next time someone want to make a change here its clear what format should be used (which is hard for a one-time contributor if the code is inconsistent).

drivers/mysql.go
svanharmelen: I disagree. The other method (`GetRecords`) expects an `offset` and a `limit` and so `totalRecords` is the result of doing the same query but then without the `offset` and `limit` (`SELECT COUNT(*) FROM TABLE`) so that the caller can see/know how many records there are in total which may not the amount of records that are returned.

The `ExecuteQuery` method on the other hand just executes a given query and returns the result. So in this case the returned number of records is just the number of records that you are going to receive. Even if you've send a query which included an `offset` and a `limit`. So calling it `totalRecords` here as well is a bit misleading in my opinion.

drivers/mssqql.go
svanharmelen: Did you read the body of the function? The code is actually really easy to read, so if you go up about 30 lines it should all explain itself.

In short, `results` contains all the rows for the result table so this includes a row with the column names.

drivers/driver.go
ccoVeille: I would say I'm agree, but I still disagree.

By adding 1 returned value, things are becoming more complicated. 

A method returning something, plus an error is classic. No need to use named return, everyone understand.

The returned value you add is an interface, but it's unclear what it could be.

I can understand you disagree. It's OK. The same way you disagree with my suggestions.

I don't know if other interfaces are something that I consider unclear.

Now what, I don't think I would be able to use a named return here.

So I have a suggestion. Feel free to apply or ignore.

You could add a comment on the interface to explain the method returns.

If I didn't manage to convince you nevermind... I might open a PR later, or never.

All these was a point of detail.

drivers/mssqql.go
ccoVeille: Indeed, so it's the header. Got it

ccoVeille: The fact the code is clear don't make obvious why at the end you decrement the number of results.

Maybe you could :
- add a comment
- define a variable where it's about the header, so here do the -1, then return this variable at the end

Apparently, I fail to convince you but I keep trying

drivers/mysql.go
ccoVeille: You are right.

My misunderstanding is indeed linked to my other comment about the -1

drivers/driver.go
svanharmelen: >By adding 1 returned value, things are becoming more complicated.

While that's certainly true, the `GetRecords` method, for example, also has 3 return params. So I'm still not adding a new pattern to this interface in which case I usually try to follow the existing style of the existing code to keep things consistent.

I guess the same goes for adding a comment. While I agree that helps to better understand the methods in this interface, at that same time I think it should then be done consistently to all methods which (in my opinion) is out of scope of this PR as it's a change that is completely not related to what I am trying to fix here.

drivers/mssqql.go
svanharmelen: I think we can do something better here to improve readability.... Will push an update in a minute.

ccoVeille (approved):

PR #154, opened 2025-01-13
Partial fix to delete row funcionality

### LINKED ISSUE #153: Uncheck Delete focus on row
If I'm selecting few rows to delete. It shows Red bg(which is nice) but I want to uncheck some rows before committing changes. how do i do that? 
And a bug. If i switch tabs( by pressing 1,2, 3 ). I lose that Red bg.
switching between editor and table ( by pressing [] ) is good( Red bg remains )

Thanks!

jorgerojas26:
This is a partial fix to be able to mark and unmark rows to be deleted and keep row coloring.

But i am aware there are a lot of problems to keep the state of the changes being made to the rows (update, delete, append).

I will make a huge revamp to that logic in another PR.

Fixes #153

### DIFF
--- a/components/results_table.go
+++ b/components/results_table.go
@@ -301,6 +301,10 @@ func (table *ResultsTable) AppendNewRow(cells []models.CellValue, index int, UUI
 	for i, cell := range cells {
 		tableCell := tview.NewTableCell(cell.Value.(string))
 		tableCell.SetExpansion(1)
+		// Appended rows have a reference to the row UUID so we can identify them later
+		// Also, rows that have columns marked to be UPDATED will have a reference to the type of the new value (NULL, EMPTY, DEFAULT)
+		// So, the cell reference will be used to determine if the row/column is an inserted row or if it's an UPDATED row
+		// I'm sure there is a better way to do this, but it works for now
 		tableCell.SetReference(UUID)
 		tableCell.SetTextColor(app.Styles.PrimaryTextColor)
 		tableCell.SetBackgroundColor(tcell.ColorDarkGreen)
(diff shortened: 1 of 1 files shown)


ccoVeille:
Typo spotted: functionality, not funcionality

components/results_table.go
```diff
@@ -301,6 +301,10 @@ func (table *ResultsTable) AppendNewRow(cells []models.CellValue, index int, UUI
 	for i, cell := range cells {
 		tableCell := tview.NewTableCell(cell.Value.(string))
 		tableCell.SetExpansion(1)
+		// Appended rows have a reference to the row UUID so we can identify them later
+		// Also, rows that have columns marked to be UPDATED will have a reference to the type of the new value (NULL, EMPTY, DEFAULT)
+		// So, the cell reference will be used to determine if the row/column is an inserted row or if it's an UPDATED row
+		// I'm sure there is a better way to do this, but it works for now
```
ccoVeille, line 307: Usually comment like this are not written at first person. Using the passive voice may help


```suggestion
		// there might be a better way to do this, but it works for now
```

components/results_table.go
```diff
@@ -1375,3 +1372,36 @@ func (table *ResultsTable) UpdateSidebar() {
...
+				isAnInsertedRow := cellReference != nil && cellReference.(string) != "NULL&" && cellReference.(string) != "EMPTY&" && cellReference.(string) != "DEFAULT&"
+
+				if isAnInsertedRow {
+					return true, i
+				}
+			}
+		}
+
+	}
+	return false, -1
```
ccoVeille, line 1393: I would like to suggest


```suggestion
func (table *ResultsTable) isAnInsertedRow(rowIndex int) (isAnInsertedRow bool, DBChangeIndex int) {
	for i, dmlChange := range *table.state.listOfDBChanges {

		for _, value := range dmlChange.Values {
			if value.TableRowIndex != rowIndex {
				continue
			}

			cellReference := table.GetCell(rowIndex, 0).GetReference()
			if cellReference == nil {
				continue
			}

			switch cellReference.(string) {
				 case "NULL&", "EMPTY&", "DEFAULT&":
					continue
				default:
					return true, i
			}
		}
	}
	return false, -1
```

this is pseudo code written on my phone, please check it

ccoVeille: But maybe this would be more efficient

```suggestion
func (table *ResultsTable) isAnInsertedRow(rowIndex int) (isAnInsertedRow bool, DBChangeIndex int) {
	for i, dmlChange := range *table.state.listOfDBChanges {

		for _, value := range dmlChange.Values {
			if value.TableRowIndex != rowIndex {
				continue
			}

			cellReference := table.GetCell(rowIndex, 0).GetReference()
			if cellReference == nil {
				break
			}

			switch cellReference.(string) {
				 case "NULL&", "EMPTY&", "DEFAULT&":
				default:
					return true, i
			}
			break
		}
	}
	return false, -1
```

PR #162, opened 2025-02-09
Update README.md

tkolleh:
* The Spanish phrase is more to the point, and it's a great way to share phrases from other languages with the English-dominated tech world.


jorgerojas26:
Hahaha, this is good.

PR #163, opened 2025-02-10
feat: add offset layout to the sidebar

### LINKED ISSUE #160: Sidebar overlay behaviour
The current behaviour of the (right) sidebar where it overlays the table can be a bit disorienting, especially so when the cell cursor disappears behind it. I can't even see the last column on some tables without hiding it.

Would it be possible to change it to instead behave like the left navigation sidebar? Or at least not overlay the table to the point where I can't see what I'm doing.

Really enjoying lazysql, thank you very much for all the work you put into it.

jorgerojas26:
The main idea of this PR is to add an offset layout to the sidebar, so users can choose between offset and overlay layout as they want with the new `SidebarOverlay` config option.

Also, the default value for `SidebarOverlay` is false, so the offset layout is the default now.

Offset layout:
image

Overlay layout:
image

Fixes #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ URL = 'postgres://postgres:urlencodedpassword@localhost:5432/foo'
 [application]
 DefaultPageSize = 300
 DisableSidebar = false
+SidebarOverlay = false
 ```
 
 The `[aplication]` section is used to define some app settings. Not all settings are available yet, this is a work in progress.
--- a/app/config.go
+++ b/app/config.go
@@ -21,6 +21,7 @@ func defaultConfig() *Config {
 	return &Config{
 		AppConfig: &models.AppConfig{
 			DefaultPageSize: 300,
+			SidebarOverlay:  false,
 		},
 	}
 }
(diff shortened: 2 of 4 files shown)


syhol:
Thanks so much @jorgerojas26 🎉 🙏

svanharmelen:
Very nice feature @jorgerojas26 👍🏻 

I just updated and enabled the sidebar again, but I noticed that the lines of the table "overflow" while the records summary and where clause boxes do not (they look the same as in the snapshot you posted). Is that a bug, of should I set any additional config to fix that?

image

svanharmelen:
Very nice feature @jorgerojas26 👍🏻 

I just updated and enabled the sidebar again, but I noticed that the lines of the table "overflow" while the records summary and where clause boxes do not (they look the same as in the snapshot you posted). Is that a bug, of should I set any additional config to fix that?

Normal example:
image

Highlighted example:
image

svanharmelen:
Ah check, I understand...

PR #173, opened 2025-03-13
Fix broken query when reserved name is used, fix GUID showing

kmiterror:
So I really like this project.
I found few issues with MS SQL db.
First of all I couldn't open databases with name like User 
It seems that if the word is "reserved" then it becomes a problem 
You can see it even in the Query editor 
this will fail:
 `select * from user`
 
 this will work:
 `select * from [user]`
PS  In the query editor, it's impossible to enter [  and ] characters.
 It will take you to another tab.
 I don't know how to fix it yet.
 
 Also I noticed that GUIDS are scrambled, this not only looks bad but also cripples sorting mechanism (pressing J or K to sort the column)
 So 2 birds 1 stone situation.

### DIFF
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -10,6 +10,7 @@ import (
 	_ "github.com/microsoft/go-mssqldb"
 	"github.com/xo/dburl"
 
+	"github.com/google/uuid"
 	"github.com/jorgerojas26/lazysql/helpers/logger"
 	"github.com/jorgerojas26/lazysql/models"
 )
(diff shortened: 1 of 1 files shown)


kmiterror:
I see that you fixed the reserved name issue,
I will close this PR and create another one with just GUID fix.

ccoVeille:
> I will close this PR and create another one with just GUID fix.

For reference:
- #174 

> I see that you fixed the reserved name issue

I assume it was fixed with 0262df44ee9027f68c2c039d992b17148eb96a71 with #170

But I'm perplex about the code currently in mssql.go

I think the select are OK, but I'm unsure about the DELETE and UPDATE

https://github.com/jorgerojas26/lazysql/blob/aac446b482838cdf8289d40d83f1b80a750d7718/drivers/mssqql.go#L332

https://github.com/jorgerojas26/lazysql/blob/aac446b482838cdf8289d40d83f1b80a750d7718/drivers/mssqql.go#L361

@jorgerojas26 could you take a look?

jorgerojas26:
> > I will close this PR and create another one with just GUID fix.
> 
> For reference:
> 
> * [show guids properly in MS SQL #174](https://github.com/jorgerojas26/lazysql/pull/174)
> 
> > I see that you fixed the reserved name issue
> 
> I assume it was fixed with [0262df4](https://github.com/jorgerojas26/lazysql/commit/0262df44ee9027f68c2c039d992b17148eb96a71) with #170
> 
> But I'm perplex about the code currently in mssql.go
> 
> I think the select are OK, but I'm unsure about the DELETE and UPDATE
> 
> https://github.com/jorgerojas26/lazysql/blob/aac446b482838cdf8289d40d83f1b80a750d7718/drivers/mssqql.go#L332
> 
> https://github.com/jorgerojas26/lazysql/blob/aac446b482838cdf8289d40d83f1b80a750d7718/drivers/mssqql.go#L361
> 
> @jorgerojas26 could you take a look?

Hi, those "update" and "delete" methods in fact are not being used in the code. Everything was moved to the "ExecutePendingChanges". So, they need to be deleted on all drivers and removed from the driver.go file

PR #174, opened 2025-03-14
show guids properly in MS SQL

kmiterror:
Guids are not shown properly in the table
This fixes it
Before
image

After
image

### DIFF
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -10,6 +10,7 @@ import (
 	_ "github.com/microsoft/go-mssqldb"
 	"github.com/xo/dburl"
 
+	"github.com/google/uuid"
 	"github.com/jorgerojas26/lazysql/helpers/logger"
 	"github.com/jorgerojas26/lazysql/models"
 )
@@ -244,7 +245,7 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
 		query += fmt.Sprintf(" %s", where)
 	}
 
-	// since in mssql order is mandatory when using pagination
+	// Since in MSSQL, ORDER BY is mandatory when using pagination
 	if sort == "" {
 		sort = "(SELECT NULL)"
 	}
@@ -275,19 +276,36 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
 		if err := rows.Scan(rowValues...); err != nil {
 			return nil, 0, err
 		}
-
+		// Get column types to identify UNIQUEIDENTIFIER
+		columnTypes, err := rows.ColumnTypes()
+		if err != nil {
+			return nil, 0, err
+		}
 		var row []string
-		for _, col := range rowValues {
+		for i, col := range rowValues {
 			if col == nil {
 				row = append(row, "NULL&")
 				continue
 			}
 
-			colval := string(*col.(*sql.RawBytes))
-			if colval == "" {
-				row = append(row, "EMPTY&")
+			rawBytes := *col.(*sql.RawBytes)
+			colType := columnTypes[i].DatabaseTypeName()
+
+			if colType == "UNIQUEIDENTIFIER" {
+				// Convert binary GUID to standard string format
+				guid, err := uuid.FromBytes(rawBytes)
+				if err != nil {
+					return nil, 0, fmt.Errorf("failed to parse GUID: %w", err)
+				}
+				row = append(row, guid.String())
 			} else {
-				row = append(row, string(*col.(*sql.RawBytes)))
+				// Handle non-GUID columns as strings
+				colval := string(rawBytes)
+				if colval == "" {
+					row = append(row, "EMPTY&")
+				} else {
+					row = append(row, colval)
+				}
 			}
 		}
 
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -8,10 +8,10 @@ import (
 
 	"github.com/google/uuid"
 	_ "github.com/microsoft/go-mssqldb"
+	"github.com/xo/dburl"
 
 	"github.com/jorgerojas26/lazysql/helpers/logger"
 	"github.com/jorgerojas26/lazysql/models"
-	"github.com/xo/dburl"
 )
 
 type MSSQL struct {
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -4,12 +4,11 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"github.com/google/uuid"
 	"strings"
-
 	// mssql driver - register MSSQL driver with database/sql
 	_ "github.com/microsoft/go-mssqldb"
 
-	"github.com/google/uuid"
 	"github.com/jorgerojas26/lazysql/helpers/logger"
 	"github.com/jorgerojas26/lazysql/models"
 	"github.com/xo/dburl"
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -4,12 +4,12 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
-	"github.com/google/uuid"
 	"strings"
 
 	// mssql driver - register MSSQL driver with database/sql
 	_ "github.com/microsoft/go-mssqldb"
 
+	"github.com/google/uuid"
 	"github.com/jorgerojas26/lazysql/helpers/logger"
 	"github.com/jorgerojas26/lazysql/models"
 	"github.com/xo/dburl"
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -303,16 +303,16 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
 					return nil, 0, fmt.Errorf("failed to parse GUID: %w", err)
 				}
 				row = append(row, guid.String())
-			} else {
-				// Handle other columns as strings
-				colval := string(*rawBytes)
-				if colval == "" {
-					colval = "EMPTY&"
-				}
-				row = append(row, colval)
+				continue
 			}
-		}
 
+			// Handle other columns as strings
+			colval := string(*rawBytes)
+			if colval == "" {
+				colval = "EMPTY&"
+			}
+			row = append(row, colval)
+		}
 		results = append(results, row)
 	}
 
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -304,7 +304,7 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
 				} else {
 					// Fallback to hex string if parsing fails
 					hexValue := hex.EncodeToString(*rawBytes)
-					row = append(row, "0x123"+hexValue) // Prefix with "0x" for clarity
+					row = append(row, "0x"+hexValue) // Prefix with "0x" for clarity
 					logger.Warn("Invalid GUID", map[string]any{
 						"table":  table,
 						"column": columns[i],
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/google/uuid"
+	// ms sql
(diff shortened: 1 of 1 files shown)


drivers/mssqql.go
```diff
@@ -275,19 +276,36 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
+			colType := columnTypes[i].DatabaseTypeName()
+
+			if colType == "UNIQUEIDENTIFIER" {
+				// Convert binary GUID to standard string format
+				guid, err := uuid.FromBytes(rawBytes)
+				if err != nil {
+					return nil, 0, fmt.Errorf("failed to parse GUID: %w", err)
+				}
+				row = append(row, guid.String())
 			} else {
```
ccoVeille, line 301: You can easily remove the else

```suggestion
			if colType == "UNIQUEIDENTIFIER" {
				// Convert binary GUID to standard string format
				guid, err := uuid.FromBytes(rawBytes)
				if err != nil {
					return nil, 0, fmt.Errorf("failed to parse GUID: %w", err)
				}
				row = append(row, guid.String())
				continue
 			}
```

drivers/mssqql.go
```diff
@@ -275,19 +276,36 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
-				row = append(row, string(*col.(*sql.RawBytes)))
+				// Handle non-GUID columns as strings
+				colval := string(rawBytes)
+				if colval == "" {
+					row = append(row, "EMPTY&")
+				} else {
+					row = append(row, colval)
+				}
```
ccoVeille, line 308: This code could be simplified, also the comment about non-GUID is overkill here, so here is my suggestion

```suggestion
				// Handle other columns as strings
				colval := string(rawBytes)
				if colval == "" {
					colVal = "EMPTY&"
				}
				row = append(row, colval)
```

drivers/mssqql.go
```diff
@@ -275,19 +276,36 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
+			rawBytes := *col.(*sql.RawBytes)
+			colType := columnTypes[i].DatabaseTypeName()
+
+			if colType == "UNIQUEIDENTIFIER" {
+				// Convert binary GUID to standard string format
+				guid, err := uuid.FromBytes(rawBytes)
+				if err != nil {
+					return nil, 0, fmt.Errorf("failed to parse GUID: %w", err)
+				}
+				row = append(row, guid.String())
```
ccoVeille, line 300: Are we sure all unique identifiers are all GUID?

I don't know mssql enough, but can a unique identifier be something different than a valid GUID ?

I mean does it worth blocking everything for one that might not pass.

Maybe using the hex version, instead of returning an error

https://pkg.go.dev/encoding/hex#EncodeToString

drivers/mssqql.go
```diff
@@ -10,6 +10,7 @@ import (
 	_ "github.com/microsoft/go-mssqldb"
 	"github.com/xo/dburl"
 
+	"github.com/google/uuid"
```
ccoVeille, line 13: Please fix the imports order

```suggestion
	"github.com/google/uuid
	// mssql driver
 	_ "github.com/microsoft/go-mssqldb"
	"github.com/xo/dburl""
```

drivers/mssqql.go
```diff
@@ -275,19 +276,36 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
 			if col == nil {
 				row = append(row, "NULL&")
 				continue
 			}
 
-			colval := string(*col.(*sql.RawBytes))
-			if colval == "" {
-				row = append(row, "EMPTY&")
+			rawBytes := *col.(*sql.RawBytes)
+			colType := columnTypes[i].DatabaseTypeName()
```
ccoVeille, line 292: I would add security checks here


```suggestion
			rawBytes, ok := *col.(*sql.RawBytes)
			if !ok {
				return nil, 0, errors.New("unexpected whatever")
			}
			columnType, ok := columnTypes[i]
			if !ok {
				return nil, 0, errors.New("unexpected whatever")
			}
			colType := columnType.DatabaseTypeName()
```

drivers/mssqql.go
```diff
@@ -7,9 +7,10 @@ import (
 	"strings"
 
 	// mssql driver
```
ccoVeille, line 9: Please remove this one

```suggestion
```

I wasn't able to include it in the suggestion I made

drivers/mssqql.go
```diff
@@ -275,19 +276,42 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
-			colval := string(*col.(*sql.RawBytes))
-			if colval == "" {
-				row = append(row, "EMPTY&")
+			rawBytes, ok := *col.(*sql.RawBytes)
+			if !ok {
+				return nil, 0, errors.New("unexpected whatever")
+			}
+			columnType, ok := columnTypes[i]
+			if !ok {
+				return nil, 0, errors.New("unexpected whatever")
```
ccoVeille, line 297: The whatever here had to be adapted 😅

drivers/mssqql.go
jorgerojas26: Yeah, i don't think it is a good idea to return an error only for that.

drivers/mssqql.go
kmiterror: I tried this suggestion but I got those errors
Can you help out?
I started like yesterday with go and I don't know how to fix this.

image

kmiterror: Ok, nevermind I fixed that one

jorgerojas26:
Hahahaha this is funny, don't let the ci win the fight.

kmiterror:
> Hahahaha this is funny, don't let the ci win the fight.

Sorry for the chaos
I have no idea what am I doing 
I started with go like yesterday
Just added some lsp to my neovim setup, maybe this will make things easier
Stay tuned for more fun :D

jorgerojas26:
> > Hahahaha this is funny, don't let the ci win the fight.
> 
> Sorry for the chaos I have no idea what am I doing I started with go like yesterday Just added some lsp to my neovim setup, maybe this will make things easier Stay tuned for more fun :D

That's okay, i'm not experienced in Go either.

You can run the linter locally with the command "golangci-lint run".

drivers/mssqql.go
kmiterror: I tried using hex.EncodeToString but this outputs data in weird format
it looks like this
`0xe02ef36a2cf479478c48a366fd5e4ddd`

this is not very readable, most users will rather enjoy this format
`e02ef36a-2cf4-7947-8c48-a366fd5e4ddd`

What do you guys think?

ccoVeille: I'm sorry if you misunderstood my comment

My idea was to consider using the hex format when the guid decoding fails.

So instead of returning on error, you fallback on the hex version then.

So it's only about not returning an error when guide fails, but to use another format in such case

I hope it's clearer.

drivers/mssqql.go
```diff
@@ -275,22 +277,54 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
+
+			columnType := columnTypes[i]
+			colType := columnType.DatabaseTypeName()
+
+			if colType == "UNIQUEIDENTIFIER" {
+				// Try to parse as a GUID
+				if guid, err := uuid.FromBytes(*rawBytes); err == nil {
+					row = append(row, guid.String()) // Use standard GUID format if valid
+				} else {
+					// Fallback to hex string if parsing fails
```
kmiterror, line 305: @ccoVeille please have a look once more

drivers/mssqql.go
```diff
@@ -275,22 +277,54 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
+					})
+				}
+				continue
+			}
+
+			// Handle other columns as strings
+			colval := string(*rawBytes)
+			// Check nullability and handle empty strings
+			nullable, _ := columnType.Nullable()
+			if nullable && colval == "" {
```
ccoVeille, line 322: Good idea, but while it will work for the use case you thought about, I don't think it will work for field that are empty when the field is nullable.

Please make test with a table containing a nuallable string with the following cases

NULL
""
"foobar"

But maybe the if statement catching the `if col == nil` make it work

drivers/mssqql.go
```diff
@@ -2,11 +2,13 @@ package drivers
...
 import (
 	"database/sql"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"strings"
 
-	// mssql driver
+	"github.com/google/uuid"
+	// ms sql
```
ccoVeille, line 11:
```suggestion
	// MSSQL driver
```

drivers/mssqql.go
```diff
@@ -275,22 +277,54 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
 		if err := rows.Scan(rowValues...); err != nil {
 			return nil, 0, err
 		}
-
+		// Get column types to identify UNIQUEIDENTIFIER
+		columnTypes, err := rows.ColumnTypes()
+		if err != nil {
+			return nil, 0, err
+		}
```
ccoVeille, line 284: I'm paranoid, but to avoid panic when accessing columnTypes[i] a few lines below, I would add this


```suggestion
		columnTypes, err := rows.ColumnTypes()
		if err != nil {
			return nil, 0, err
		}

		if len(columnTypes) != len(rowValues) {
			return nil, 0, errors.New("unexported number of column")
		}
```

I dislike panic

This was what I reported here

https://github.com/jorgerojas26/lazysql/pull/174#discussion_r1995484157

But my suggestion wasn't working 

```go
columnType, ok := columnTypes[i]
if !ok {
}
```

This one works for maps, not slices

drivers/mssqql.go
```diff
@@ -275,22 +277,54 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
+			colType := columnType.DatabaseTypeName()
+
+			if colType == "UNIQUEIDENTIFIER" {
+				// Try to parse as a GUID
+				if guid, err := uuid.FromBytes(*rawBytes); err == nil {
+					row = append(row, guid.String()) // Use standard GUID format if valid
+				} else {
+					// Fallback to hex string if parsing fails
+					hexValue := hex.EncodeToString(*rawBytes)
+					row = append(row, "0x123"+hexValue) // Prefix with "0x" for clarity
```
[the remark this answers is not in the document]
kmiterror: ahh, leftover from some testing
good catch

drivers/mssqql.go
kmiterror: NULL
image

""
image

"GETUTCDATE()"
image

For me it looks exactly the same way as it's presented in other sql editors
Is there something that I'm missing?

ccoVeille: Thanks for checking

kmiterror: Is there anythig else to check here?

drivers/mssqql.go
```diff
@@ -275,22 +277,58 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
...
 			return nil, 0, err
 		}
+		// Get column types to identify UNIQUEIDENTIFIER
+		columnTypes, err := rows.ColumnTypes()
+		if err != nil {
+			return nil, 0, err
+		}
 
+		if len(columnTypes) != len(rowValues) {
+			return nil, 0, errors.New("unexported number of column")
```
ccoVeille, line 287: oups, my bad I introduced a typo


```suggestion
			return nil, 0, errors.New("unexpected number of column")
```

drivers/mssqql.go
ccoVeille: No, it's fine

ccoVeille (approved):
LGTM 👍, except a typo (I made)

ccoVeille (approved):

PR #175, opened 2025-03-14
[PROPOSITION] make it possible to jump by half-page (aka 5 items) instead of 1 item

kmiterror:
ctrl+u page up
ctrl+d means page down


jorgerojas26:
I was thinking about this the other day hehe, thanks for implementing it.

PR #176, opened 2025-03-14
change keybind from [] to ctrl+[] for navigation between tabs

kmiterror:
This is a breaking change
The issue is that query editor doesn't allow to enter [] characters because they are bound to tab navigation
I'm not sure if this is the best way to fix it but it works and seems sensible.


kmiterror:
@jorgerojas26 @ccoVeille  what do you guys think about that change?

jorgerojas26:
I have a bit of muscle memory in [], but i totally understand this, i think it is needed. I'm going to merge.

PR #186, opened 2025-04-02
feat: Lazysql can be embedded in applications that already have a sqlite driver registered

jbleyaert:
When trying to use lazysql inside an existing project, I run into 
```
panic: sql: Register called twice for driver sqlite
```

This is because modernc runs the init() function when imported.
The problem is that the other project needs the import to use gorm. Lazysql only imports it to use the sqlite driver.
Adding this build tag would fix this issue.

Open to suggestions.

### DIFF
--- a/drivers/sqlite.go
+++ b/drivers/sqlite.go
@@ -6,9 +6,6 @@ import (
 	"fmt"
 	"strings"
 
-	// import sqlite driver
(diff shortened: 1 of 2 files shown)


jbleyaert:
@jorgerojas26 Could you kindly review this PR?

Issue #156, opened 2025-01-20
[feat] populate password for `command` output


a1994sc:
I have a postgresql connection that I would like to look at, but it has a randomly generated password.

The config looks like this:

```toml
[[database]]
Name = 'cluster-coder'
Provider = 'postgres'
DBName = 'coder'
URL = 'postgres://postgres:<password>@localhost:${port}/coder?sslmode=disable'
Commands = [
  { Command = 'kubectl --context cluster port-forward -n coder svc/cluster-pooler ${port}:5432', WaitForPort = '${port}' }
]
```

It would be nice to be able to populate the `<password>` the same we do for `${port}`.

```toml
Commands = [
  { Command = 'kubectl --context cluster get secrets blah | yq .data.password | base64 -d', <password identifier> },
  { Command = 'kubectl --context cluster port-forward -n coder svc/cluster-pooler ${port}:5432', WaitForPort = '${port}' }
]
```

I'm not sure what it should look like for `<password identifier>` to say I want the output of this command to be used as the password... this might be able to be extend to allow arbitrary variable substitution, but at least MVP for this issue would be able to populate the password dynamically.

jorgerojas26:
@svanharmelen Hi mate, i hope you are doing well. You are the one who did the commands feature, would you mind to check this and maybe help with a PR?

svanharmelen:
Hi @jorgerojas26 I'm super busy at the moment, so don't really have time to look into this one...

svanharmelen:
@jorgerojas26 and @a1994sc I think I can come up with a solution, just not sure in what time frame as I'm a bit overloaded at the moment. But will keep this one in the back of my head to take a look at later when things have settled down again...

a1994sc:
@svanharmelen please work at your own place, this is a nice to have not a requirement so take your time

CalebJohnHunt:
I've started working on this problem. Specifically, I'm adding custom variables. I don't want to step on any toes if others are getting close though.

Is other work nearly there? Or should I keep working on my implementation?

Also, just to bring the discussion back, here's what I'm thinking for the custom variables:
```
# config.toml
[[database]]
...
URL = "driver://me:${password}@..."
Commands = [
  { Command = "password-getter", CreatesTheFollowingVariable = "password" },
  { Command = "echo Here's my password: ${password}" },
]
```
(The second command isn't useful; it's just showing off capabilities.)

[...]

PR #197, opened 2025-06-12
Implement user-define variables

CalebJohnHunt:
Implements: #156

Any thoughts on the `SaveOutputTo` field name?

I'm also not in love with the `waitToCaptureVariable` solution. I think it would make more sense for the `helpers.RunCommand` to return a `<- chan string` with the output, and then we could `onDone := App.Register(); go func() { _ = <- c; onDone() }()`. In other words, `RunCommand` doesn't take in the `onDone` callback; it just gives out a channel which we can use to synchronize. I didn't feel like it was important enough to figure it out here though.  
Similarly, I'm unhappy with the `io.MultiWriter` usage. It feels a bit hacky to me.

Also, in case you're curious, here's my original use case for this feature: I have two MSSQL instances on my machine. If you have just one, it's typically on `localhost:1433`, but with two, their ports are actually dynamic. At time of writing, my connection string is `sqlserver://user:pass@localhost:47904`, but that port will randomize on restart! So I wrote a powershell script to get it. Now I set up my config.toml like
```toml
[[database]]
# ...
URL = 'sqlserver://user:pass@localhost:${dynamicPort}'
Commands = [
  { Command = 'powershell.exe get-port.ps1', SaveOutputTo = 'dynamicPort' }
]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,9 +152,10 @@ The configuration file is a TOML file and can be used to define multiple connect
 Name = 'Production database'
 Provider = 'postgres'
 DBName = 'foo'
-URL = 'postgres://postgres:urlencodedpassword@localhost:${port}/foo'
+URL = 'postgres://${user}:urlencodedpassword@localhost:${port}/foo'
 Commands = [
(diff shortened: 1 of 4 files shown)


jorgerojas26:
I think this is good.

PR #215, opened 2025-09-19
feat: add DuplicateRow command

nytoan:
Hello,

I added a functionality that i needed and was missing. It's the ability to **duplicate a row** (just under an existing row) already filled with the same values.

I hope it can be merged and help some other users !


jorgerojas26:
This is awesome, thank you.

PR #216, opened 2025-09-24
MSSQL: fix the way GUIDs are shown

kmiterror:
Currently guids are shown like:
39887ea6-d811-c64c-8611-eefb36ad631e 
but should be
a67e8839-11d8-4cc6-8611-eefb36ad631e
note how the order is flipped in eg. c64c and 4cc6

This is how it looks in current lazysql version:
image

This PR fixes it:
image

Now it's same as the one shown in other MS SQL editors, eg. Visual Studio Code below:
image

### DIFF
--- a/drivers/mssqql.go
+++ b/drivers/mssqql.go
@@ -21,6 +21,35 @@ type MSSQL struct {
 	Provider   string
 }
 
+// mssqlGUIDToUUID converts a 16-byte little-endian GUID from MSSQL
+// into a standard uuid.UUID.
+func mssqlGUIDToUUID(dbBytes []byte) (uuid.UUID, error) {
+	if len(dbBytes) != 16 {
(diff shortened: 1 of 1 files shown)


drivers/mssqql.go
```diff
@@ -307,8 +336,10 @@ func (db *MSSQL) GetRecords(database, table, where, sort string, offset, limit i
 
 			if colType == "UNIQUEIDENTIFIER" {
 				// Try to parse as a GUID
-				if guid, errParse := uuid.FromBytes(*rawBytes); errParse == nil {
-					row = append(row, guid.String()) // Use standard GUID format if valid
+				// if guid, errParse := uuid.FromBytes(*rawBytes); errParse == nil {
+				// 	row = append(row, guid.String()) // Use standard GUID format if valid
```
ccoVeille, line 340: This looks like left behind code

PR #217, opened 2025-09-29
Add complex commands

nytoan:
When using Command in the config file, the command doesn't support operators like |, &&, >.

If i want to do something like that :

`{ Command = "kubectl get secrets -n namespace my-sql-credentials -o jsonpath={.data.PGPASSWORD} | base64 -d | jq -sRr @uri", SaveOutputTo = 'password' }, `

it throw a parse error because there is a pipe.

The solution is to run the command with a shell.

### DIFF
--- a/helpers/command.go
+++ b/helpers/command.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"os"
 	"os/exec"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


jorgerojas26:
why don't just use a shell as part of the command?

nytoan:
Hi Jorge,

Sadly It does not work using a shell because of the `parts := strings.Fields(command)`. Using a shell requires to be able to give all the command in one string.

jorgerojas26:
Thank you guys for your time and your help making lazysql better. I'm going to close this PR.

PR #229, opened 2025-10-25
Toggle left wrapper tree with modal

itsmewes:
This is an initial attempt at giving an option to move the db tree list and search to a modal instead of a left panel based on a config setting. I think there is still some work to be done, for example I don't like the config name at this point and I'm sure the code is a bit messy and could be optimised and the modal hangs on nested trees like postgres if the left wrapper is not added and the tui hangs if launching lazysql with the database string eg. `lazysql mysql://...`, but just want to check if you are happy with the direction, otherwise I'll just keep this as part of my fork.

To get it to work add `HideTableTree = true` to the application part of the config

This PR:
- Creates a way to show/hide the left panel via a `F` keyboard shortcut, though I don't think in a good way since I need the panel to be added to the maincontent for the modal to be able to be shown.
- adds a new table list modal that takes in the tree.
- Adds a new config that hides the left panel and uses the modal instead. So keyboard shortcuts like `Ctrl + P` still work.
- You can still go to the left panel using the same `H` shortcut.

### DIFF
--- a/components/results_table.go
+++ b/components/results_table.go
@@ -257,7 +257,10 @@ func (table *ResultsTable) subscribeToSidebarChanges() {
 			}
 
 			logger.Info("eventSidebarCommitEditing", map[string]any{"cellValue": cellValue, "params": params, "rowIndex": row, "changedColumnIndex": changedColumnIndex})
-			table.AppendNewChange(models.DMLUpdateType, row, changedColumnIndex, cellValue)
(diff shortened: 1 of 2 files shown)


itsmewes:
> I like this, let me just take a deeper look.
> 
> Thank you very much for contribution

Amazing, thank you for taking a deeper look.

itsmewes:
Closing this PR in favour of #248

PR #232, opened 2025-11-03
Add programming features

iihmsunn:
Allows listing and opening of stored procedures, user defined functions and views, if the driver reports that it supports it. Only MSSQL implementation is currently included.

Additional changes:
- The code that handles selecting the tree nodes is rewritten to be more explicit about what exactly is being selected.
- Driver now also reports if it's using schemas, that is if you have to select both database and a schema before you see the tables (and only postgres one currently does), which allowed to remove some postgres hardcoded logic from the tree code.
- MSSQL driver now properly passes the selected database into queries. Previously only one (default) database would be used for most if not all queries.


iihmsunn:
Some of them I don't currently know how to fix, specifically iirc you can't properly pass an argument to an mssql query if it's not a value but a database name. I'll look into it a bit more when I have time, maybe you can put it somewhere outside the query or something.

PR #234, opened 2025-11-09
feat: support making read only connections

byuly:
implements a read-only mode for database connections to prevent accidental data mutations.

## summary of changes: 
 - added `ReadOnly` boolean field to the `Connection` model
 - checkbox in connection form UI to mark connections as read-only when making them
 - read-only setting persist with TOML configuration file
 - cannot toggle at runtime, fixed when creating
 - still supports statements that DON'T modify the original table (SELECT, SHOW, TMP VIEW, ETC.)
 - cell editing also disabled, as well as keybinds
 - executing statements that excuete modifying queries will make a popup that connection its currently in read only.
 - new test suite for SQL mutation detection (`drivers/validation_test.go`)
 - read only connection can also be made with --read-only flag in args

 
## UI changes:
- display read-only in light blue color in tree panel and results table
- READ prefix when selecting a connection, to indicate that the connection is read only

PS. also changed logic for detecting SELECT flow, found while testing
 - query routing logic that incorrectly used `Contains()` instead of `HasPrefix()`, which caused
  DELETE/UPDATE queries with SELECT subqueries to bypass validation 
ex. "DELETE FROM products WHERE product_id IN (SELECT product_id FROM products WHERE price < 10);" will result in SELECT flow.

## supporting images:
 
image
image
image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ Name = 'Production database'
 Provider = 'postgres'
 DBName = 'foo'
 URL = 'postgres://${user}:urlencodedpassword@localhost:${port}/foo'
+ReadOnly = true
 Commands = [
   { Command = 'ssh -tt remote-bastion -L ${port}:localhost:5432', WaitForPort = '${port}' },
   { Command = 'whoami', SaveOutputTo = 'user' },
@@ -166,6 +167,8 @@ DefaultPageSize = 300
 DisableSidebar = false
 SidebarOverlay = false
 ```
+
+The `ReadOnly` field (optional, defaults to `false`) can be set to `true` to enable read-only mode for a connection. When enabled, all mutation queries (INSERT, UPDATE, DELETE, DROP, etc.) will be blocked.
 
 The `[application]` section is used to define some app settings. Not all settings are available yet, this is a work in progress.
 
@@ -184,6 +187,12 @@ $ lazysql [connection_url]
 ```
 
 To launch lazysql and connect to database at [connection_url].
+
+```console
+$ lazysql --read-only [connection_url]
+```
+
+To launch lazysql in read-only mode.
 
 ### Connect to a DB
 


components/results_table.go
```diff
@@ -832,7 +876,7 @@ func (table *ResultsTable) SetError(err string, done func()) {
 	table.state.error = err
 
 	table.Error.SetText(err)
-	table.Error.SetDoneFunc(func(_ int, _ string) {
+	table.Error.SetDoneFunc(func(buttonIndex int, buttonLabel string) {
```
jorgerojas26: I think this change is not needed, variables are unused

components/home.go
```diff
@@ -349,6 +357,16 @@ func (home *Home) homeInputCapture(event *tcell.EventKey) *tcell.EventKey {
...
 		}
 	case commands.Save:
+		if home.ReadOnly {
+			errorModal := tview.NewModal().
+				SetText("Cannot save changes: Connection is in read-only mode").
+				AddButtons([]string{"OK"}).
+				SetDoneFunc(func(buttonIndex int, buttonLabel string) {
+					mainPages.RemovePage("readOnlyError")
+				})
+			mainPages.AddPage("readOnlyError", errorModal, true, true)
```
jorgerojas26: page names are being put in `/components/constants.go`

byuly:
thanks for the review, addressed them!

jorgerojas26:
Hi @byuly. Thank you for addressing my suggestions.

I think this is good for merge, but then i just thought about the command `lazysql [connection_url]`. I think we should support a flag like "--read-only" to be able to directly connect to a database in readonly mode.

What do you think, we can merge this now and implement that flag in another PR, or maybe implement it here.

Let me know what you think about.

And, thank you very much for this PR, i really appreciate the help.

byuly:
was a simple add, so added onto this PR. and no worries @jorgerojas26 !

PR #264, opened 2026-01-22
Add optional timeout parameter to commands

super-cooper:
External commands run for a given database connection may now have an
optional `TimeoutSeconds` parameter included, giving users the ability
to change the timeout duration from the default 5 seconds.

Update to the README includes a practical example for fetching a database password from a
password manager, which is likely to block while the user enters their
master password


jorgerojas26:
Closed in favor of #233

PR #293, opened 2026-04-15
feat: Confirmation pop-up when accidentally press "q"

Liferenko:
I'm sure I'm not the only one who's losing it after accidentally misclicking "q"
I've no idea where this muscle-memory move comes from. It is definitely not from nvim, nor from CLI etc etc. But the fact is - I press it constantly, subconsciously, and it is annoying. 

I'd be happy if it helps somebody else.


jorgerojas26:
I have suffer that too hehe...

I think this is a good idea, but maybe we can do it better.

From a UI perspective, i think it would be better to have that confirmation window a bit smaller and compact.

Also, it would be good to have shortcuts on both "yes" and "no" options, if i press "y" is yes, if i press "n" is no, so i don't have to press 2 buttons to say "no". I know you can do "Esc" and close the window and that would count as "no", but i think it's better UX.

What do you think about?

Thanks for your contribution.

Liferenko:
totally make sense. I'll do y/n confirmation

stcrd:
i suffer both from pressing q and 'ctrl-c', either one closes lazysql immediately

will this confirmation be shown on all 'quit' commands?

Liferenko:
@stcrd , good catch. I added ctrl+c handler as well. 

@jorgerojas26 , for now both `q` and `ctrl+c` end up with confirmation window. Enter/Esc, y/n, Y/N work as expected

PR #322, opened 2026-07-12
feat(drivers): add MongoDB driver and NoSQL driver interface

byuly:
MongoDB/NoSQL doesn't fit the existing SQL Driver interface if you look at what its methods actually expect. GetRecords returns [][]string, basically a grid of table cells, and takes where and sort as raw string fragments that get built into a SQL query. Mongo documents aren't tabular though, they're nested BSON with embedded documents and arrays, so flattening one into a []string row loses structure, and there's no WHERE clause to build a string from since Mongo queries are bson.M documents, not query language text.

A chunk of the interface exists purely to build parameterized SQL strings and describe stored procedures and views, none of which Mongo has, since it talks typed documents over its own driver instead of building query strings at all.

So NoSQLDriver has its own Document, Filter, Sort, and Schema types that actually match how Mongo represents data instead of forcing it through methods built around SQL strings and table rows. GetSchema works by sampling up to 100 documents and inferring field types from what it finds, since there's no fixed schema to read from a catalog. Filtering uses a small FilterCondition type with operators like eq, ne, gt, lt, in, nin, contains, and regex that map onto bson query operators.

There's also a small factory in the drivers package that builds a driver by provider name, for both SQL and NoSQL. Right now nothing calls the NoSQL side of it, the SQL callers still have their own switch statements,  and next PR will clean it up. Wanted the factory to live here so there's one obvious place to register future NoSQL databases later without touching UI code.

Kept the interface minimal for now, just connect and read operations. No insert, update, or delete yet since nothing in the UI would call them, didn't want half wired code sitting around.


jorgerojas26:
Hi @byuly 

I really like this approach more. Thank you for doing this. I think this PR is great.

byuly:
> Hi @byuly
> 
> I really like this approach more. Thank you for doing this. I think this PR is great.

Hi @jorgerojas26,

Of course! Let me know if you have any questions, I spent some time splitting it up into digestible pieces. I'll put them up one by one after merges. This is the first part with the Drivers factory refactor + MongoDB driver integration.

PR #329, opened 2026-08-04
fix(mssql): quote database names with FormatReference to support special characters

jova1111:
#### Summary
Fixed showing database tables, functions, procedures and views for mssql databases in the tree when database name contains dashes.

#### Problems

- Database names containing dashes (e.g. 'mydatabase-tst') were concatenated directly into SQL strings without bracket quoting, causing SQL Server to parse them as arithmetic expressions and fail to show children in the tree.

- `addProgrammingNodes` used `node.GetText()` to look up database-name keys
in the procedures/functions/views maps. That method returns modified database name if node is selected in the tree (eg. color info is appended). Map lookup will fail if the user navigates to the database node in the tree before DB info is fetched.
#### Changes

- replaced all raw database name interpolation with `db.FormatReference(database)`
- replaced table/column/key names to use the same method
- updated tests for mssql driver to reflect this change
- using `node.GetReference() to get clean database name as a key for programming nodes dictionaries


jorgerojas26:
Closed in favor of #343

PR #332, opened 2026-08-05
fix: scope tree search results to database when using dotted paths (banco.schema.tabela)

phellipw:
## What

Fixes tree search so that typing a qualified path like `banco.tabela` or
`banco.schema.tabela` actually scopes results to that database (and schema,
for the 3-part form), instead of leaking matches from every other database
in the connected instance.

## The bug (as reported)

> mesmo eu colocando banco.schema.tabela ele filtra de todos, não só do
> banco que estou procurando

`Tree.search()` already had support for a *two-part, space-separated*
qualified search (e.g. `"auth users"` → schema `auth`, table `users`), but
it only ever split the query on a single space:

```go
parts := strings.SplitN(lowerSearchText, " ", 2)
```

Typing `banco.schema.tabela` has no space in it, so the entire dotted
string — dots included — was treated as one opaque `tableNameFilter`, with
`databaseNameFilter` left empty. That falls into the "no database filter"
branch, which fuzzy-matches the literal string against **every node in the
tree**, regardless of which database it lives under. Any table whose name
happened to fuzzy-match the dotted string (which fuzzy matching is
generous about) would show up, from any database.

## Fix

- Extracted parsing into `parseSearchQuery(query) (ancestorFilters []string, tableNameFilter string)`:
  - `"users"` → no ancestor filters, table filter `users`
  - `"db.users"` → ancestor filters `[db]`, table filter `users`
  - `"db.schema.users"` → ancestor filters `[db, schema]`, table filter `users`
  - `"schema users"` (legacy space syntax) → ancestor filters `[schema]`, table filter `users` — unchanged
  - Dots take priority over spaces when a query contains both.
- Generalized the ancestor-matching walk in `search()` from a single
  hardcoded database filter to an **ordered list** of ancestor filters.
  Each filter must match a distinct ancestor of the candidate node, walked
  outermost-to-innermost (through section headers like
  tables/views/functions, reusing the chain-walk this file already had for
  the space-based two-part case). **A node is only considered a match if
  every ancestor filter is satisfied** — this is the actual fix: it's what
  scopes `dbA.users` to dbA instead of also matching dbB's `users` table.

## Testing

- `TestParseSearchQuery`: parser edge cases (dots vs. spaces, stray/
  duplicate dots, dot priority over space, dots-only input).
- `TestSearch_DottedTwoPartScopesToDatabase` / `..._OtherDatabaseScopesCorrectly`:
  direct regression test for the reported bug — two databases each with a
  same-named `users` table; `dbA.users` returns only dbA's node, `dbB.users`
  returns only dbB's.
- `TestSearch_DottedThreePartScopesToDatabaseAndSchema`: same table name
  under two schemas in one database; `db.schemaA.users` disambiguates
  correctly.
- `TestSearch_SpaceSeparatedTwoPartStillWorks`: regression guard for the
  pre-existing space-based syntax.
- All pre-existing search tests (`TestSearch_TwoPartFindsDeepNodeThroughSectionHeaders`,
  `TestSearch_SinglePartWorksNormally`, `TestSearchRanking_*`) pass
  unmodified.
- `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- `gofmt -l .` clean.

## Risks / rollback

- Purely additive parsing change; single-part search behavior is byte-for-
  byte identical to before. The legacy space-separated two-part syntax is
  preserved unchanged and covered by regression tests.
- Known limitation (pre-existing, not introduced by this change): database/
  schema/table identifiers containing a literal `.` or space in their name
  aren't disambiguated from the separator — same class of limitation the
  space-based syntax already had for names containing spaces.
- Contained to `components/tree.go`'s `search`/new `parseSearchQuery`
  helper and its tests; no driver or query execution code touched.


phellipw:
## Increment: exact-match collapse for fully-qualified queries

Following up on user feedback: the ancestor-scoping fix above correctly stops cross-database/schema leaking, but the final segment was still fuzzy-matched even when the full qualified path (`db.schema.table`) matched an existing node exactly. Near-name tables in the same schema (e.g. `dad` vs `dado`) could still both show up.

**Fix:** once at least one ancestor filter is present (dotted or legacy space syntax) and an exact case-insensitive match exists among the ancestor-scoped candidates, the result set now collapses to just that exact match — near-name fuzzy siblings are dropped. Partial/mid-typing qualified queries with no exact hit yet keep fuzzy-matching unchanged (narrow-as-you-type UX preserved). Unqualified single-part search is untouched.

New tests: `TestSearch_DottedThreePartExactMatchCollapsesNearNameSiblings`, `TestSearch_DottedThreePartPartialSegmentStillFuzzyMatches`, `TestSearch_SpaceSeparatedExactMatchCollapsesNearNameSiblings`, `TestSearch_SinglePartExactMatchDoesNotCollapseSiblings`. All 8 pre-existing search tests pass unmodified. `go build`, `go vet`, `go test ./...`, `gofmt -l .` all clean.

Issue #333, opened 2026-08-07
table titles are not visible when there are 5+ tables tab open


Liferenko:
Also when `[` `]`-switching tabs - you don't see latest tabs' title
current tab is visible:
<img width="1240" height="47" alt="Image" src="https://github.com/user-attachments/assets/c0a2e588-5a5d-4c4a-b258-14f3a304af55" />

current tab is not visible:
<img width="1241" height="44" alt="Image" src="https://github.com/user-attachments/assets/1076cd6e-8221-44d3-8269-c65c580697de" />

zoom out does slightly resolve the issue. But it is a tempo solution
zoomed out:
<img width="1348" height="20" alt="Image" src="https://github.com/user-attachments/assets/5eef52e1-165a-4c3b-8a1e-fcc6d5afba26" />

Liferenko:
I'll try to fix it myself and send a PR

Liferenko:
```
GIVEN a user is keeping open 10+ tabs of tables
  AND the user's laptop monitor is pretty average: 14 inch Macbook Pro
WHEN the user opens one more table
  AND new tab has been opened at the end of tab menu
THEN the user sees 11th tab opened, highlighted (secondary color)
  AND the 11th tab's name is fully visible
  AND the user doesn't need to zoom out TUI to see the 11th tab
```

Liferenko:
#341

PR #341, opened 2026-09-05
[bugfix] Broken focus on active table tab when 5+ tabs open

Liferenko:
The issue: #333 

expected behaviour:
```
GIVEN a user is keeping open 10+ tabs of tables
  AND the user's laptop monitor is pretty average: 14 inch Macbook Pro
WHEN the user opens one more table
  AND new tab has been opened at the end of tab menu
THEN the user sees 11th tab opened, highlighted (secondary color)
  AND the 11th tab's name is fully visible
  AND the user doesn't need to zoom out TUI to see the 11th tab
```

#### demo "before/after":

https://github.com/user-attachments/assets/844a3a81-baac-4c9c-8792-b328f537f0b5

### DIFF
--- a/components/tabbed_menu.go
+++ b/components/tabbed_menu.go
@@ -31,11 +31,15 @@ type TabbedPaneState struct {
 	Length     int
 }
 
+const headerArrowMaxWidth = 3
+
 type TabbedPane struct {
 	*tview.Pages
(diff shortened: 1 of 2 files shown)


Liferenko:
this bugfix tested only with PG DB tables

PR #346, opened 2026-09-14
feat: improve network performance and progressive loading

jorgerojas26:
## Summary

- Improve first-use network performance with one-page Records fetches, progressive tree/autocomplete metadata loading, surface-aware refreshes, metadata caching, and bounded row counts.
- Add context-aware, cancelable driver I/O with connection-pool controls across SQLite, MySQL, PostgreSQL, and MSSQL.
- Stream SQL results and CSV exports safely with row caps, replay-safety checks, cancellation, and atomic export handling.
- Add MySQL foreign-key metadata fast paths with compatibility fallbacks, schema-qualified MSSQL handling, structured performance logging, RTT benchmarks, manual database fixtures, and documentation.

## CI lint follow-up

- Fixed all reported `errcheck`, `gosec`, `misspell`, `revive`, `rowserrcheck`, `staticcheck`, and `unused` findings.
- Standardized cancellation terminology and made cleanup/error handling explicit.

## Validation

- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --timeout=5m`
- `go test -v ./...`
- `go build ./...`
- `git diff --check`

### DIFF
--- a/.beans.yml
+++ b/.beans.yml
@@ -1,0 +1,6 @@
+beans:
+    path: .beans
+    prefix: lazysql-
+    id_length: 4
+    default_status: todo
+    default_type: task
--- a/.beans/lazysql-532r--06-eliminate-the-mysql-340-foreign-key-metadata-ho.md
+++ b/.beans/lazysql-532r--06-eliminate-the-mysql-340-foreign-key-metadata-ho.md
@@ -1,7 +1,7 @@
 ---
 # lazysql-532r
 title: '[06] Eliminate the MySQL #340 foreign-key metadata hotspot'
-status: todo
+status: completed
 type: bug
 priority: high
 tags:
@@ -14,7 +14,7 @@ tags:
     - ready-for-agent
     - ticket-key-mysql-foreign-key-hotspot
 created_at: 2026-09-12T03:19:44Z
-updated_at: 2026-09-12T03:19:45Z
+updated_at: 2026-09-12T09:46:15Z
 parent: lazysql-a9lf
 blocked_by:
     - lazysql-846w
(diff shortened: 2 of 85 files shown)


jorgerojas26:
## Code review and hardening completed

Pushed fixes through **2716fff** and reconciled current `main` (**fd5ea57**, including PR #330). GitHub reports the PR as **MERGEABLE** (no merge conflicts).

### Findings fixed

- **P1 — streaming cleanup:** a failed consumer/CSV write could hang while the database reader blocked in `Next`. Concurrent `Rows.Close`/`Scan` also raced. Streams now own a cancelable context and join the single reader/closer before releasing the connection.
- **P1 — unsafe export replay:** executable MySQL/MariaDB comments, quoted variable assignments, qualified functions, and dialect-ambiguous syntax could bypass replay checks. Added conservative rejection and regression tests while keeping normal SELECT execution independent of export eligibility.
- **P1 — PostgreSQL write routing:** pending edits could target the login database instead of the captured database. Transactions now use the selected catalog; mixed-database batches fail before dispatch.
- **P2 — CSV data loss/races:** raw query values such as `NULL&`, `EMPTY&`, and `DEFAULT&` were treated as display markers. Raw query exports now preserve them, and visible rows are snapshotted on the UI loop.
- **P2 — progressive loading:** moved automatic count setup onto the UI loop; blocked stale bulk metadata publication after refresh/DDL; preserved cancellation ownership for schema loads.
- **Integration:** adapted ClickHouse to cancelable pooled I/O, page lookahead, counts, and typed streaming. Reverse Foreign Key Jump stays lazy/cached. Integrated PR #330 so editor streams and export replay retain the original selected database.
- **Fixture safety:** replaced committed passwords with generated, ignored, mode-0600 local files; restricted published ports to localhost. Fixed MSSQL startup recovery being mistaken for an unseeded database, which could otherwise attempt to overwrite manual fixture data.

### Verification

[CI run 35696342637](https://github.com/jorgerojas26/lazysql/actions/runs/35696342637) **passed** lint, tests, race detection, performance contracts, and build on 2716fff.

Also passed locally: three repeated full race-enabled suites; RTT benchmark matrix; real MySQL, PostgreSQL, MSSQL, and SQLite fixture tests; cross-database read/write/stream isolation; ClickHouse typed/capped streaming integration; and an MSSQL restart persistence check. Test containers have been stopped.

### Remaining external gate

**GitGuardian still fails on historical occurrences**, not newly generated credentials:

- Development-fixture password incidents: **37224298, 37224299, 37224300**, from `bddd342`.
- Interpolation false positive: **37513126**, matching the Compose required-variable error-message expression in `c75d799`; the expression was simplified in `0e100a9`.

The owner needs to review these incidents in GitGuardian (and rotate credentials if they were ever reused outside disposable fixtures). No security check was disabled and no history was force-rewritten. **Code findings are fixed and validated, but I would not call the PR fully merge-ready until this security gate is resolved.**

